PR #2, opened 2024-09-05
State parsing MVP

seanmcgary:


internal/eigenState/avsOperators.go
```diff
@@ -0,0 +1,324 @@
...
+
+// Write the new final state to the database.
+//
+// 1. Get latest distinct change value for each avs/operator
+// 2. Join the latest unique change value with the previous blocks state to overlay new changes
+// 3. Filter joined set on registered = false to get unregistrations
+// 4. Determine which rows from the previous block should be carried over and which shouldnt (i.e. deregistrations)
+// 5. Geneate the final state by unioning the carryover and the new registrations
+// 6. Insert the final state into the registered_avs_operators table
+func (a *AvsOperators) WriteFinalState(blockNumber uint64) error {
```
scotthconner: are you writing every value for every slot at every block? I'm confused whiy this is so big

internal/eigenState/avsOperators.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// Generates a state root for the given block number.
+//
+// 1. Select all registered_avs_operators for the given block number ordered by avs and operator asc
+// 2. Create an ordered map, with AVSs at the top level that point to an ordered map of operators and block numbers
+// 3. Create a merkle tree for each AVS, with the operator:block_number pairs as leaves
+// 4. Create a merkle tree for all AVS trees
+// 5. Return the root of the full tree
+func (a *AvsOperators) GenerateStateRoot(blockNumber uint64) (StateRoot, error) {
```
scotthconner: I was expectniig the state root to be a hash on the DIFF in a per block, it almost looks like with this you are producing the state root since genesis?

scotthconner: if that is the case producing the state root will take longer and longer. I expected the state root to just be a dumb hash of each table/row change or something simple like that?

seanmcgary: Can definitely switch to doing just the diffs, not a major change at all.

PR #48, opened 2024-09-13
lint and check, cleanup idiomatic go issues

jac18281828:
This cleans up many issues with idiomatic go, it also adds a requirement that these checks are enabled in the continuous build


Makefile
```diff
@@ -46,9 +52,39 @@ docker-buildx-self:
...
+		echo "$$unformatted_files"; \
+		echo "Please run 'gofmt -w .' to format them."; \
+		exit 1; \
+	fi
+.PHONY: vet
+vet:
+	go vet ./...
+
+.PHONY: lint
+lint:
```
seanmcgary, line 78: @jac18281828 you may need to update from master. I added similar targets for lint and formatting yesterday

https://github.com/Layr-Labs/go-sidecar/blob/master/Makefile#L63-L71

Makefile
```diff
@@ -46,9 +52,39 @@ docker-buildx-self:
...
 test:
 	TESTING=true go test -v -p 1 -parallel 1 ./...
 
+.PHONY: staticcheck
+staticcheck:
+	staticcheck ./...
+
 .PHONY: ci-test
-ci-test: test
+ci-test: test lint vet fmtcheck yamlfmt
```
seanmcgary, line 90: Suggest leaving these as all separate; that way we can run individual github actions for them to clearly show which is failing (also means they can run in parallel)

internal/eigenState/stateManager/stateManager_test.go
```diff
@@ -80,4 +81,6 @@ func Test_StateManager(t *testing.T) {
 		assert.Equal(t, insertedStateRoots[0].EthBlockHash, root.EthBlockHash)
 		assert.Equal(t, insertedStateRoots[0].StateRoot, root.StateRoot)
 	})
+
+	teardown(NewEigenStateManager(l, grm))
```
seanmcgary: Rather than taking a new instance of the struct, `teardown()` should probably just take a `*gorm.DB` pointer since in this context it's used to delete contents from the table used in the tests.

seanmcgary:
```go
func teardown(grm *gorm.DB) {
    grm.Exec("delete from state_roots")
}

**Review by seanmcgary — CHANGES_REQUESTED** [review]

File: .golangci.yml [hunk 4]
```diff
@@ -1,82 +1,6 @@
-
```
jac18281828, line 1: The linters below are not passing and will require some refactor to get them passing.  I backed it out to a basic setup so the ci-cd will be passing for now.    We need to take a look at the functions that have high complexity and make sure we are following the best practices there.

Makefile
jac18281828: Okay I have merged those changes.  I'd ask that we start to use approved pull requests as the way to merge so that we can all stay current on what changes are happening.

Makefile
jac18281828: yeah we can split that out into separate actions, they are run individually in the test section above.  That is probably sufficient.   This line is the dependency requirement to build the image so they all should be here.

internal/eigenState/stateManager/stateManager_test.go
jac18281828: ✅

.golangci.yml
seanmcgary: Ok sounds good! I knew they were a bit aggressive which was why I hadnt enforced it in CI yet, but backing it out is fine to get it running.

seanmcgary (approved):

PR #53, opened 2024-09-18
feat: implement cli argument parsing

### LINKED ISSUE #35: Add proper CLI interface to improve user experience for running the sidecar
Started on the branch [`sm-cli`](https://github.com/Layr-Labs/go-sidecar/tree/sm-cli) using [https://github.com/urfave/cli](urfave/cli) as the base since it seems to be less bad than Cobra/Viper.

As a first pass, the CLI just needs a single run target that can take flags. Usage should look like:

```bash
sidecar run \
    --chain "<mainnet | holesky |preprod, required>" \
    --ethereum.rpc-base-url "<string, required>" \
    --ethereum.ws-url "<string, required>" \
    --sqlite.in-memory "<true | false, default false>" \
    --sqlite.db-file-path "<string, default ./sqlite.db>" \
    --rpc.grpc-port "<number, default 7100>" \
    --rpc.http-port "<number, default 7101>" \
    --etherscan-api-keys "<string, required>"
```

_Notes_:

* `--chain` would replace the two distinct `environment` and `network` flags we have now
* Each flag should also be mapped to be accepted as an environment variable. Those are already defined in the `config/config.go` but it needs to be wired up properly.
* Precedence: Flags should take precedence over ENV vars
* Note `--etherscan-api-keys` needs to be deprecated at some point since we shouldnt take a hard dependency on it. Its mostly unused at the moment but hasnt been fully removed.

jac18281828:
closes #35 

this implements the command line argument and environment handling for the sidecar binary


internal/config/cli.go
```diff
@@ -0,0 +1,163 @@
...
+				return fmt.Errorf("--network is required: %v", err)
+			}
+			if opts.Network == "mainnet" && opts.Environment == "" {
+				opts.Environment = "mainnet"
+			}
+			if _, err := parseEnvironment(opts.Environment); err != nil {
+				return fmt.Errorf("--environment is required: %v", err)
+			} else if opts.Network == "mainnet" && opts.Environment != "mainnet" {
+				return fmt.Errorf("mainnet environment required")
+			}
```
seanmcgary: One thing I mentioned in the original ticket was that we should do away with the network/environment pairing in favor of something like `--chain=<mainnet, holesky, preprod>`.

Environment/network has been a confusing concept for people using our other CLIs like the root updater and has really only been used internally, so I think a single flag to denote the target chain makes that much easier for people to understand.

jac18281828: Yeah that does make sense!  This also is a nice simplification.  It is much more clear now.

PR #62, opened 2024-10-20
renaming

gpsanant:
small renaming changes and fixing a bug in deleteState of rewardsSubmission model

### DIFF
--- a/internal/eigenState/avsOperators/avsOperators.go
+++ b/internal/eigenState/avsOperators/avsOperators.go
@@ -72,8 +72,8 @@ type AvsOperatorsModel struct {
 	deltaAccumulator map[uint64][]*AvsOperatorStateChange
 }
 
-// NewAvsOperatorsModelModel creates a new AvsOperatorsModel.
-func NewAvsOperatorsModelModel(
+// NewAvsOperatorsModel creates a new AvsOperatorsModel.
+func NewAvsOperatorsModel(
 	esm *stateManager.EigenStateManager,
 	grm *gorm.DB,
 	logger *zap.Logger,
--- a/internal/eigenState/avsOperators/avsOperators.go
+++ b/internal/eigenState/avsOperators/avsOperators.go
@@ -72,8 +72,8 @@ type AvsOperatorsModel struct {
 	deltaAccumulator map[uint64][]*AvsOperatorStateChange
 }
 
-// NewAvsOperators creates a new AvsOperatorsModel.
-func NewAvsOperators(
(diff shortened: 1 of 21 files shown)


internal/eigenState/avsOperators/avsOperators.go
```diff
@@ -194,12 +194,18 @@ func (a *AvsOperatorsModel) IsInterestingLog(log *storage.TransactionLog) bool {
...
 }
 
-func (a *AvsOperatorsModel) InitBlockProcessing(blockNumber uint64) error {
+func (a *AvsOperatorsModel) InitBlock(blockNumber uint64) error {
 	a.stateAccumulator[blockNumber] = make(map[types.SlotID]*AccumulatedStateChange)
 	a.deltaAccumulator[blockNumber] = make([]*AvsOperatorStateChange, 0)
 	return nil
 }
 
+func (a *AvsOperatorsModel) CleanupBlock(blockNumber uint64) error {
```
seanmcgary: Rather than `InitBlock` and `ClearBlock` we should go with something like `SetupStateForBlock` and `CleanupProcessedStateForBlock`.

internal/eigenState/submittedDistributionRoots/submittedDistributionRoots.go
```diff
@@ -22,7 +22,7 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-type SubmittedDistributionRoots struct {
+type SubmittedDistributionRoot struct {
```
seanmcgary, line 25: If we wanna drop the `s` let's rename the folder/file to match

seanmcgary (changes requested):

internal/eigenState/avsOperators/avsOperators.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/62/commits/1e801ba5d5aeeabac725c43e01ac8f703b1e5265

prefer the clean names, but ok :)

internal/eigenState/submittedDistributionRoots/submittedDistributionRoots.go
gpsanant: do we? for example, `avsOperators.go` centers around a `AvsOperatorStateChange` struct. the model itself is still `SubmittedDistributionRootsModel`

seanmcgary: Ah I see what you're thinking now. This is fine the way you have it

seanmcgary (approved):

PR #63, opened 2024-10-20
feat: move HandleStateChange to base

gpsanant:
This PR aims to remove duplication between models in their log handling so the code is more readable and easier to reason about.

- Move log handlers for state transitions to base class
- Made `StateTransitions` a non generic to play nicer with new abstraction. this seems ok since return value is not used
- HandleStateChange -> HandleLog

### DIFF
--- a/internal/eigenState/avsOperators/avsOperators.go
+++ b/internal/eigenState/avsOperators/avsOperators.go
@@ -60,10 +60,9 @@ func NewSlotID(avs string, operator string) types.SlotID {
 // EigenState model for AVS operators that implements IEigenStateModel.
 type AvsOperatorsModel struct {
 	base.BaseEigenState
-	StateTransitions types.StateTransitions[AccumulatedStateChange]
-	DB               *gorm.DB
-	logger           *zap.Logger
-	globalConfig     *config.Config
+	DB           *gorm.DB
+	logger       *zap.Logger
+	globalConfig *config.Config
 
 	// Accumulates state changes for SlotIds, grouped by block number
 	stateAccumulator map[uint64]map[types.SlotID]*AccumulatedStateChange
@@ -80,12 +79,10 @@ func NewAvsOperatorsModel(
 	globalConfig *config.Config,
 ) (*AvsOperatorsModel, error) {
 	s := &AvsOperatorsModel{
-		BaseEigenState: base.BaseEigenState{
-			Logger: logger,
-		},
-		DB:           grm,
-		logger:       logger,
-		globalConfig: globalConfig,
+		BaseEigenState: base.NewBaseEigenState(logger, grm),
+		DB:             grm,
+		logger:         logger,
+		globalConfig:   globalConfig,
 
 		stateAccumulator: make(map[uint64]map[types.SlotID]*AccumulatedStateChange),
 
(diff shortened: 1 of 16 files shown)


seanmcgary:
Most of this is the exact opposite of what idiomatic Go is supposed to be. "BaseClasses" are not a thing, and explicit implementation of an interface, even if it ends up slightly more verbose, is favored over trying to simulate polymorphism.

Moving `HandleStateChange` out of each implementation assumes that all will ever be the same always, which may not be true, in which case you'll end up reverting it anyway. This change also loses types on the return values.

internal/eigenState/base/baseEigenState.go
```diff
@@ -71,13 +80,42 @@ func (b *BaseEigenState) IsInterestingLog(contractsEvents map[string][]string, l
 	return false
 }
 
-func (b *BaseEigenState) DeleteState(tableName string, startBlockNumber uint64, endBlockNumber uint64, db *gorm.DB) error {
+// HandleLog handles state changes for a given block number.
+// it loops through the forkBlockNumbers and calls the state change function for the block number
+// this supports models that have different state changes for different block ranges
+func (b *BaseEigenState) HandleLog(stateTransitions types.StateTransitions, log *storage.TransactionLog) (interface{}, error) {
```
seanmcgary, line 86: I mentioned this in the top level comment, but pushing this function into a "base class" is against idiomatic Go patterns. It assumes that all `HandleLog` functions will always be the same, which may not be true, in which case you're back to the original way this was written since Go doesnt have a mechanism for overriding a method implementation like polymorphic languages do.

You're also losing the typing of `types.StateTransitions` which is model-specific.

internal/eigenState/avsOperators/avsOperators.go
```diff
@@ -106,11 +102,11 @@ func (a *AvsOperatorsModel) GetModelName() string {
...
 // Returns the map and a reverse sorted list of block numbers that can be traversed when
 // processing a log to determine which state change to apply.
-func (a *AvsOperatorsModel) GetStateTransitions() (types.StateTransitions[AccumulatedStateChange], []uint64) {
-	stateChanges := make(types.StateTransitions[AccumulatedStateChange])
+func (a *AvsOperatorsModel) GetStateTransitions() types.StateTransitions {
+	stateTransitions := make(types.StateTransitions)
 
 	// TODO(seanmcgary): make this not a closure so this function doesnt get big an messy...
-	stateChanges[0] = func(log *storage.TransactionLog) (*AccumulatedStateChange, error) {
+	stateTransitions[0] = func(log *storage.TransactionLog) (interface{}, error) {
```
seanmcgary, line 109: Why remove the typed return?

internal/eigenState/avsOperators/avsOperators.go
```diff
@@ -80,12 +78,10 @@ func NewAvsOperatorsModel(
 	globalConfig *config.Config,
 ) (*AvsOperatorsModel, error) {
 	s := &AvsOperatorsModel{
-		BaseEigenState: base.BaseEigenState{
-			Logger: logger,
-		},
-		DB:           grm,
-		logger:       logger,
-		globalConfig: globalConfig,
+		BaseEigenState: base.NewBaseEigenState(logger, grm),
```
seanmcgary, line 81: BaseEigenState is meant to be stateless. The only reason it really exists is to define an interface. If it didnt need to take a logger, then each of it's member functions would be pure, stateless functions. Arguably, that can be done by passing the logger to each function, which is actually more in line with idiomatic Go.

internal/eigenState/base/baseEigenState.go
seanmcgary: A good alternative to not break the correct styling is to have some pure helper functions that can DRY up the contents of this function. Then you get a few benefits:

* You keep the correct idiomatic Go implementation
* You achieve your alleged goal of simplifying this code
* This function can be implemented in any way in any model without having to refactor a lot down the line when something needs to break from the existing pattern

seanmcgary (changes requested):
Base on the comments left, this change is not a positive one and needs more/better justification for its inclusion. PR should state what the actual goal of this change aims to achieve.

internal/eigenState/base/baseEigenState.go
gpsanant: Yeah I actually prefer this. 

In response to the ask for justification for the change, it's  simply to avoid repetition to make code easier to read and reason about.

If we agree on implementing a helper library I'll go ahead and do that.

internal/eigenState/avsOperators/avsOperators.go
gpsanant: This is done to make it play nice with BaseEigenState. If we end up using a pure helper method, then we can make it generic again, however the typed return value is also beacons untyped in HandleLog anyway

PR #65, opened 2024-10-21
WIP: add withdrawals

gpsanant:
Adding the withdrawals model. This is critical protocol data that needs to be kept track of for serving data to customers via RPC. This does not include RPC endpoints.

Notably this
- Adds bindings to the eigenState directory
- - how to avoid this?
- Adds blocktimes to the logs and transaction tables


internal/sqlite/migrations/202409061249_bootstrapDb/up.go
```diff
@@ -21,6 +21,7 @@ func (m *SqliteMigration) Up(grm *gorm.DB) error {
     	)`,
 		`CREATE TABLE IF NOT EXISTS transactions (
 			block_number INTEGER NOT NULL REFERENCES blocks(number) ON DELETE CASCADE,
+			block_time DATETIME NOT NULL,
```
seanmcgary, line 24: Since migrations are only run once, this needs to go in a brand new migration as an `alter table` statement

PR #68, opened 2024-10-29
Gpsanant/slashing events

gpsanant:
Adds support for slashing processing by updating the `staker_shares`

This keeps track of a single `diffAccumulator` for the model which stores both share deltas (due to deposits and withdrawals) and slashings.

When `prepareState` is called, the `deltaAccumulator` is populated by converting all slashings to deltas by looking up the shares to slash with data from the existing `staker_shares` table.

Lots of unit tests are provided, however there's a decent amount of duplication. I'm open to suggestions to dedupe!

### DIFF
--- a/pkg/eigenState/stakerShares/stakerShares.go
+++ b/pkg/eigenState/stakerShares/stakerShares.go
@@ -634,13 +634,13 @@ func (ss *StakerSharesModel) CleanupProcessedStateForBlock(blockNumber uint64) e
 }
 
 func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (interface{}, error) {
-	StateDiffs, sortedBlockNumbers := ss.GetStateTransitions()
+	stateChanges, sortedBlockNumbers := ss.GetStateTransitions()
 
 	for _, blockNumber := range sortedBlockNumbers {
 		if log.BlockNumber >= blockNumber {
 			ss.logger.Sugar().Debugw("Handling state change", zap.Uint64("blockNumber", blockNumber))
 
-			change, err := StateDiffs[blockNumber](log)
+			change, err := stateChanges[blockNumber](log)
 			if err != nil {
 				return nil, err
 			}
@@ -653,8 +653,12 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
 	return nil, nil
 }
 
+// TODO: this is not very performant in cases of large slashings and is called twice (once in CommitFinalState and once in GenerateStateRoot)
+// but prob ok?
 // prepareState prepares the state for commit by adding the new state to the existing state.
 func (ss *StakerSharesModel) prepareState(blockNumber uint64) ([]*StakerSharesDiff, error) {
+	// reset the delta accumulator
+	ss.deltaAccumulator[blockNumber] = make([]*StakerShareDeltas, 0)
 	// keep track of the updated staker shares
 	updatedStakerShares := make(map[types.SlotID]*StakerSharesDiff)
 
@@ -797,7 +801,8 @@ func (ss *StakerSharesModel) prepareState(blockNumber uint64) ([]*StakerSharesDi
 					LogIndex:        slashDiff.LogIndex,
 					BlockNumber:     slashDiff.BlockNumber,
 				})
-				// update the shares with wadsLeft
+
+				// update the shares
 				currStakerShares.Shares = currStakerShares.Shares.Add(currStakerShares.Shares, diffShares)
 				currStakerShares.BlockNumber = blockNumber
 
--- a/pkg/eigenState/stakerShares/stakerShares.go
+++ b/pkg/eigenState/stakerShares/stakerShares.go
@@ -26,27 +26,29 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-type StakerShares struct {
-	Staker      string
-	Strategy    string
-	Shares      string
-	BlockNumber uint64
-	CreatedAt   time.Time
+var WAD = big.NewInt(1e18)
+
+type StateDiffEvent interface {
+	SlotID() types.SlotID
 }
 
-type AccumulatedStateChange struct {
-	Staker      string
-	Strategy    string
-	Shares      *big.Int
-	BlockNumber uint64
+type StateDiff struct {
+	Event StateDiffEvent
 }
 
-type StakerSharesDiff struct {
-	Staker      string
-	Strategy    string
-	Shares      *big.Int
-	BlockNumber uint64
-	IsNew       bool
+type SlashDiff struct {
+	Operator        string
+	Strategy        string
+	WadsSlashed     *big.Int
+	TransactionHash string
+	LogIndex        uint64
+	BlockTime       time.Time
+	BlockDate       string
+	BlockNumber     uint64
+}
+
+func (u *SlashDiff) SlotID() types.SlotID {
+	return types.SlotID(fmt.Sprintf("MAX_MAGNITUDE_%s_%s", u.Operator, u.Strategy))
 }
 
 // Table staker_share_deltas
@@ -62,20 +64,49 @@ type StakerShareDeltas struct {
 	BlockNumber     uint64
 }
 
-func NewSlotID(staker string, strategy string) types.SlotID {
+func (ss *StakerShareDeltas) SlotID() types.SlotID {
+	return NewShareDeltaSlotID(ss.Staker, ss.Strategy)
+}
+
+func NewShareDeltaSlotID(staker, strategy string) types.SlotID {
 	return types.SlotID(fmt.Sprintf("%s_%s", staker, strategy))
+}
+
+type StakerShares struct {
+	Staker      string
+	Strategy    string
+	Shares      string
+	BlockNumber uint64
+	CreatedAt   time.Time
+}
+
+type StakerSharesDiff struct {
+	Staker      string
+	Strategy    string
+	Shares      *big.Int
+	BlockNumber uint64
+}
+
+func (ss *StakerShares) ToStakerSharesDiff() (*StakerSharesDiff, error) {
+	shares, success := numbers.NewBig257().SetString(ss.Shares, 10)
+	if !success {
+		return nil, xerrors.Errorf("Failed to convert shares to big.Int: %s", ss.Shares)
+	}
+	return &StakerSharesDiff{
+		Staker:      ss.Staker,
+		Strategy:    ss.Strategy,
+		Shares:      shares,
+		BlockNumber: ss.BlockNumber,
+	}, nil
 }
 
 type StakerSharesModel struct {
 	base.BaseEigenState
-	StateTransitions types.StateTransitions[AccumulatedStateChange]
-	DB               *gorm.DB
-	logger           *zap.Logger
-	globalConfig     *config.Config
+	DB           *gorm.DB
+	logger       *zap.Logger
+	globalConfig *config.Config
 
-	// Accumulates state changes for SlotIds, grouped by block number
-	stateAccumulator map[uint64]map[types.SlotID]*AccumulatedStateChange
-
+	diffAccumulator  map[uint64][]*StateDiff
 	deltaAccumulator map[uint64][]*StakerShareDeltas
 }
 
@@ -90,7 +121,7 @@ func NewStakerSharesModel(
 		DB:               grm,
 		logger:           logger,
 		globalConfig:     globalConfig,
-		stateAccumulator: make(map[uint64]map[types.SlotID]*AccumulatedStateChange),
+		diffAccumulator:  make(map[uint64][]*StateDiff),
 		deltaAccumulator: make(map[uint64][]*StakerShareDeltas),
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -26,27 +26,29 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-type StakerShares struct {
-	Staker      string
-	Strategy    string
-	Shares      string
-	BlockNumber uint64
-	CreatedAt   time.Time
+var WAD = big.NewInt(1e18)
```
seanmcgary: This needs to be more descriptive. WAD is pretty useless. Verbose is better than an acronym/abbreviation that leaves people guessing

seanmcgary: Also this needs to be a constant, but cant be since you're creating a new value. Recommend creating a function that returns this value to avoid potential modification side effects.

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -62,20 +64,49 @@ type StakerShareDeltas struct {
...
+}
+
+type StakerSharesDiff struct {
+	Staker      string
+	Strategy    string
+	Shares      *big.Int
+	BlockNumber uint64
+}
+
+func (ss *StakerShares) ToStakerSharesDiff() (*StakerSharesDiff, error) {
```
seanmcgary: Rather than putting this on the the StakerShares struct and coupling it to the StakerSharesDiff type, this should be a standalone function called `StakerSharesToStakerSharesDiff`

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -151,14 +182,16 @@ func (ss *StakerSharesModel) handleStakerDepositEvent(log *storage.TransactionLo
...
 
-	return &StakerShareDeltas{
-		Staker:          stakerAddress,
-		Strategy:        outputData.Strategy,
-		Shares:          shares.String(),
-		StrategyIndex:   uint64(0),
-		LogIndex:        log.LogIndex,
-		TransactionHash: log.TransactionHash,
-		BlockNumber:     log.BlockNumber,
+	return &StateDiff{
```
seanmcgary: Why? StateDiff only contains a StakerShareDeltas struct so why is it needed at all?

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -410,102 +449,120 @@ func parseLogOutputForSlashingWithdrawalQueuedEvent(outputDataStr string) (*slas
...
 		records = append(records, r)
 	}
 	return records, nil
 }
 
-type AccumulatedStateChanges struct {
-	Changes []*AccumulatedStateChange
+type operatorSlashedOutputData struct {
+	Operator string `json:"operator"`
+	// OperatorSet struct {
```
seanmcgary: Remove if not needed

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -410,102 +449,120 @@ func parseLogOutputForSlashingWithdrawalQueuedEvent(outputDataStr string) (*slas
...
+	// 	Avs           string `json:"avs"`
+	// 	OperatorSetId uint32 `json:"operatorSetId"`
+	// }
+	Strategies  []string      `json:"strategies"`
+	WadsSlashed []json.Number `json:"wadsSlashed"`
```
seanmcgary, line 460: If this can be anything bigger than a uint64 then you're gonna need to write a custom decoder to handle this, otherwise it's going to overflow.

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -26,27 +26,29 @@ import (
...
+	SlotID() types.SlotID
 }
 
-type AccumulatedStateChange struct {
-	Staker      string
-	Strategy    string
-	Shares      *big.Int
-	BlockNumber uint64
+type StateDiff struct {
+	Event StateDiffEvent
```
seanmcgary: As I got further down, I realized what you're doing, and this isnt good lol.

While this makes sense in a polymorphic language, this is not a good pattern for Go. You should be able to use generics and do something like:

```
type StateDiff[T any] struct {
    Event T
}
```
but even that isn't friendly....

The problem here is that since in Go you cant describe the contents of the struct you're trying to generically represent, you're basically treating `StateDiffEvent` as `interface{}` which isnt clear at all.

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -410,102 +449,120 @@ func parseLogOutputForSlashingWithdrawalQueuedEvent(outputDataStr string) (*slas
...
+
+	stateDiffs := make([]*StateDiff, 0)
+
+	for i, strategy := range outputData.Strategies {
+		wadsSlashed, success := numbers.NewBig257().SetString(outputData.WadsSlashed[i].String(), 10)
+		if !success {
+			return nil, xerrors.Errorf("Failed to convert wadsSlashed to big.Int: %s", outputData.WadsSlashed[i])
+		}
+		stateDiffs = append(stateDiffs, &StateDiff{
+			Event: &SlashDiff{
```
seanmcgary: See my above comment. StateDiff taking a "generic" type event is an antipattern in Go

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -610,88 +653,172 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
...
+	// keep track of the updated staker shares
+	updatedStakerShares := make(map[types.SlotID]*StakerSharesDiff)
 
-	slotIds := make([]types.SlotID, 0)
-	for slotId := range accumulatedState {
-		slotIds = append(slotIds, slotId)
-	}
+	for _, diff := range ss.diffAccumulator[blockNumber] {
+		switch diff.Event.(type) {
+		case *StakerShareDeltas:
```
seanmcgary: Reflection like this should be avoided if possible. You're better off defining a struct that can be sparsely populated with separately named and type fields. e.g. 

```go
type StateDiff struct {
	StakerShareDelta *StakerSharesDeltas
	SlashDiff *SlashDiff
}
```

Then you can do a simple if `if diff.StakerShareDelta != nil {`

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -610,88 +653,172 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
...
 
-	if res.Error != nil {
-		ss.logger.Sugar().Errorw("Failed to fetch staker_shares", zap.Error(res.Error))
-		return nil, res.Error
-	}
+			currStakerShares, ok := updatedStakerShares[slotID]
+			if !ok {
+				stakerShares := &StakerShares{}
+				// get the record from the database with the given staker and strategy with the latest blockNumber
+				res := ss.DB.Model(&StakerShares{}).Where("staker = ? AND strategy = ?", shareDelta.Staker, shareDelta.Strategy).Order("block_number desc").First(&stakerShares)
```
seanmcgary: This DB query should be put in a function just to clean this up a bit.

`func (ss *StakerSharesModel) GetStakerSharesForStakerAndStrategy(staker, strategy string, blockNumber uint64)`

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -610,88 +653,172 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
...
-					zap.Uint64("blockNumber", blockNumber),
+				return nil, xerrors.Errorf("Failed to convert shares to big.Int: %s", shareDelta.Shares)
+			}
+			currStakerShares.Shares = currStakerShares.Shares.Add(currStakerShares.Shares, shareDeltaShares)
+
+			// update the local cache
+			updatedStakerShares[slotID] = currStakerShares
+		case *SlashDiff:
+			slashDiff := diff.Event.(*SlashDiff)
+			query := `
```
seanmcgary: Same with this - put this in a function of its own to make this less big

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -610,88 +653,172 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
...
+
+			// get the staker shares for the stakers who were delegated to the operator
+			// and update the shares with the new max magnitude
+			res := ss.DB.Raw(query,
+				sql.Named("blockNumber", slashDiff.BlockNumber),
+				sql.Named("logIndex", slashDiff.LogIndex),
+				sql.Named("operator", slashDiff.Operator),
+				sql.Named("strategy", slashDiff.Strategy),
+			).Scan(&stakerShares)
+			if res.Error != nil && res.Error != gorm.ErrRecordNotFound {
```
seanmcgary: Since you're not just finding a single record, I dont think you'll ever reach an ErrRecordNotFound error. You'll just get an empty array back

pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -610,88 +653,172 @@ func (ss *StakerSharesModel) HandleStateChange(log *storage.TransactionLog) (int
...
+
+				// update the shares
+				currStakerShares.Shares = currStakerShares.Shares.Add(currStakerShares.Shares, diffShares)
+				currStakerShares.BlockNumber = blockNumber
+
+				// update the local cache
+				updatedStakerShares[slotID] = currStakerShares
 			}
-			prepared.Shares = existingShares.Add(existingShares, newState.Shares)
+		default:
```
seanmcgary: Can remove this since there is no default action

seanmcgary (changes requested):

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

gpsanant: Let me know if this pattern looks fine, resolving other comments on the same topic for now

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: just inlined it https://github.com/Layr-Labs/sidecar/pull/68/commits/849bd0db77bd553611d6ccdf776de875882dc9c5

pkg/eigenState/stakerShares/stakerShares.go
gpsanant: this one will not be bigger than 64 bits. but shares may? so how do we get away with not needing custom decoders for those?

PR #104, opened 2024-11-13
Update SlotId to be more unique

seanmcgary:
* SlotID is now based off of txHash + logIndex at a minimum
* Models that require more uniqueness add more fields as a suffix to txHash + logIndex
* Added transactionHash and logIndex columns to models that were lacking them
* Added parentHash to the blocks table


0xrajath (approved):
LGTM

PR #107, opened 2024-11-13
Add constraints to state models to reflect SlotID uniqueness

seanmcgary:


pkg/eigenState/stakerShares/stakerShares.go
```diff
@@ -46,7 +46,7 @@ type StakerShareDeltas struct {
 	WithdrawalRootString string `gorm:"-"`
 }
 
-func NewSlotID(staker string, strategy string, strategyIndex uint64, transactionHash string, logIndex uint64) types.SlotID {
+func NewSlotID(transactionHash string, logIndex uint64, staker string, strategy string, strategyIndex uint64) types.SlotID {
```
0xrajath, line 49: I think you could get away with just strategy Index right (i.e instead of strategy too)?

pkg/postgres/migrations/202411131200_eigenStateModelConstraints/up.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (m *Migration) Up(db *sql.DB, grm *gorm.DB) error {
+	queries := []string{
+		`alter table disabled_distribution_roots add constraint uniq_disabled_distribution_root unique (transaction_hash, log_index, block_number)`,
+		`alter table operator_share_deltas add constraint uniq_operator_share unique (transaction_hash, log_index, block_number, operator, staker, strategy)`,
+		`alter table reward_submissions add constraint uniq_reward_submission unique (transaction_hash, log_index, block_number, reward_hash, strategy_index)`,
+		`alter table reward_submissions drop constraint reward_submissions_reward_hash_strategy_block_number_key`,
+		`alter table staker_delegation_changes add constraint uniq_staker_delegation_change unique (transaction_hash, log_index, block_number)`,
+		`alter table staker_share_deltas add constraint uniq_staker_share_delta unique (transaction_hash, log_index, block_number, staker, strategy, strategy_index)`,
+		`alter table submitted_distribution_roots add constraint uniq_submitted_distribution_root unique (transaction_hash, log_index, block_number)`,
```
0xrajath, line 19: You actually don't need `block_number` here. 

If it's 1:1 to the log, you just need `transaction_hash` and `log_index`.

`transaction_hash` is guaranteed to never occur again in the EVM's history or infact any other EVM chain.

0xrajath: For all other models that's not 1:1 to the logs, you just need one or two additional items to distinguish the rows.

0xrajath: i.e just `strategyIndex` instead of `strategy` as well.

0xrajath: I'm pretty sure there's a performance hit to doing a compound unique constraint (i.e fewer fields means better performance). So we should strive to have just the necessary fields.

pkg/eigenState/stakerShares/stakerShares.go
seanmcgary: In theory yes, but I know for a fact this will be 100% unique. I already encountered some funky edge cases, so gonna leave it like this.

pkg/eigenState/stakerShares/stakerShares.go
0xrajath, line 1: Just double checking that all the other EigenStates also have slot ID with just txhash and logIndex?

pkg/postgres/migrations/202411131200_eigenStateModelConstraints/up.go
seanmcgary: > there's a performance hit to doing a compound unique constraint

On insert there is, but thats basically it, and we're performing (relatively speaking) so few inserts on a per-block basis that it's not going to make a noticeable difference. Thats my assumption at least, I'll profile it and see how it goes.

0xrajath (approved):

PR #117, opened 2024-11-22
feat: utilize split in existing rewards calculation

0xrajath:
Utilize the `OperatorAVSSplit` and `OperatorPISplit` in the existing rewards v1 and Programmatic incentives calculation (instead of the hardcoded 10% split).

### DIFF
--- a/pkg/rewards/2_goldStakerRewardAmounts.go
+++ b/pkg/rewards/2_goldStakerRewardAmounts.go
@@ -118,7 +118,7 @@ token_breakdowns AS (
       WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
         (sott.total_staker_operator_payout * 0.10)::text::decimal(38,0)
       ELSE
-        floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0)
+        floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
     END as operator_tokens,
     CASE
       WHEN sott.snapshot < @amazonHardforkDate AND sott.reward_submission_date < @amazonHardforkDate THEN
@@ -126,7 +126,7 @@ token_breakdowns AS (
       WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
         sott.total_staker_operator_payout - ((sott.total_staker_operator_payout * 0.10)::text::decimal(38,0))
       ELSE
-        sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0)
+        sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
     END as staker_tokens
   FROM staker_operator_total_tokens sott
   LEFT JOIN operator_avs_split_snapshots oas
--- a/pkg/rewards/2_goldStakerRewardAmounts.go
+++ b/pkg/rewards/2_goldStakerRewardAmounts.go
@@ -117,6 +117,8 @@ token_breakdowns AS (
         cast(sott.total_staker_operator_payout * 0.10 AS DECIMAL(38,0))
       WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
         (sott.total_staker_operator_payout * 0.10)::text::decimal(38,0)
+      WHEN sott.snapshot < @arnoHardforkDate AND sott.reward_submission_date < @arnoHardforkDate THEN
+        floor(sott.total_staker_operator_payout * 0.10)
       ELSE
         floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
     END as operator_tokens,
@@ -125,6 +127,8 @@ token_breakdowns AS (
         sott.total_staker_operator_payout - cast(sott.total_staker_operator_payout * 0.10 as DECIMAL(38,0))
       WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
         sott.total_staker_operator_payout - ((sott.total_staker_operator_payout * 0.10)::text::decimal(38,0))
+      WHEN sott.snapshot < @arnoHardforkDate AND sott.reward_submission_date < @arnoHardforkDate THEN
+        sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * 0.10)
       ELSE
         sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
     END as staker_tokens
@@ -145,6 +149,7 @@ func (rc *RewardsCalculator) GenerateGold2StakerRewardAmountsTable(snapshotDate 
 		zap.String("destTableName", destTableName),
 		zap.String("amazonHardforkDate", forks[config.Fork_Amazon]),
 		zap.String("nileHardforkDate", forks[config.Fork_Nile]),
+		zap.String("arnoHardforkDate", forks[config.Fork_Arno]),
 	)
 
 	query, err := renderQueryTemplate(_2_goldStakerRewardAmountsQuery, map[string]string{
@@ -159,6 +164,7 @@ func (rc *RewardsCalculator) GenerateGold2StakerRewardAmountsTable(snapshotDate 
 	res := rc.grm.Exec(query,
 		sql.Named("amazonHardforkDate", forks[config.Fork_Amazon]),
 		sql.Named("nileHardforkDate", forks[config.Fork_Nile]),
+		sql.Named("arnoHardforkDate", forks[config.Fork_Arno]),
 	)
 	if res.Error != nil {
 		rc.logger.Sugar().Errorw("Failed to create gold_staker_reward_amounts", "error", res.Error)
--- a/pkg/rewards/2_goldStakerRewardAmounts.go
+++ b/pkg/rewards/2_goldStakerRewardAmounts.go
@@ -108,15 +108,6 @@ staker_operator_total_tokens AS (
     END as total_staker_operator_payout
   FROM staker_proportion
 ),
--- Include the operator_avs_split_snapshots table
-operator_avs_splits_cte AS (
-  SELECT
-    operator,
-    avs,
-    snapshot,
-    split
-  FROM operator_avs_split_snapshots
-),
 -- Calculate the token breakdown for each (staker, operator) pair with dynamic split logic
 -- If no split is found, default to 1000 (10%)
 token_breakdowns AS (
@@ -138,7 +129,7 @@ token_breakdowns AS (
         sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0)
     END as staker_tokens
   FROM staker_operator_total_tokens sott
-  LEFT JOIN operator_avs_splits_cte oas
+  LEFT JOIN operator_avs_split_snapshots oas
   ON sott.operator = oas.operator AND sott.avs = oas.avs AND sott.snapshot = oas.snapshot
 )
 SELECT * from token_breakdowns
--- a/pkg/rewards/2_goldStakerRewardAmounts.go
+++ b/pkg/rewards/2_goldStakerRewardAmounts.go
@@ -2,6 +2,7 @@ package rewards
 
 import (
 	"database/sql"
+
 	"github.com/Layr-Labs/sidecar/internal/config"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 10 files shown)


pkg/rewards/2_goldStakerRewardAmounts.go
```diff
@@ -107,26 +108,38 @@ staker_operator_total_tokens AS (
...
     CASE
       WHEN snapshot < @amazonHardforkDate AND reward_submission_date < @amazonHardforkDate THEN
-        total_staker_operator_payout - cast(total_staker_operator_payout * 0.10 as DECIMAL(38,0))
+        total_staker_operator_payout - cast(total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0 AS DECIMAL(38, 0))
       WHEN snapshot < @nileHardforkDate AND reward_submission_date < @nileHardforkDate THEN
-        total_staker_operator_payout - ((total_staker_operator_payout * 0.10)::text::decimal(38,0))
+        total_staker_operator_payout - ((total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0)::text::decimal(38, 0))
       ELSE
-        total_staker_operator_payout - floor(total_staker_operator_payout * 0.10)
+        total_staker_operator_payout - floor(total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0)
```
ypatil12: Do we know what type the output of `COALESCE(oas.split, 1000) / 10000.0` is? We want the `total_staker_operator_payout * split` to be a `numeric * numeric` and not a `numeric * float`

pkg/rewards/5_goldRfaeStakers.go
```diff
@@ -106,12 +107,23 @@ staker_operator_total_tokens AS (
     FLOOR(staker_proportion * tokens_per_day_decimal) as total_staker_operator_payout
   FROM staker_proportion
 ),
--- Calculate the token breakdown for each (staker, operator) pair
+-- Include the operator_pi_split_snapshots table
+operator_pi_splits_cte AS (
```
ypatil12: Nit: we don't use cte in naming anywhere

ypatil12 (commented):
Looks good, one clarification

pkg/rewards/5_goldRfaeStakers.go
0xrajath: I did that to prevent a conflict with the existing `operator_pi_splits` table in the DB. Needed something to explicitly differentiate it.

pkg/rewards/2_goldStakerRewardAmounts.go
0xrajath: I mean the existing `0.10` is a float / double. So I've stuck with that. I believe the floor does round it down to a whole number.

pkg/rewards/5_goldRfaeStakers.go
seanmcgary: Not sure why this is needed at all. Why not just reference operator_pi_split_snapshots where you need it? You don’t really gain any advantage by selecting just a few columns from it

0xrajath: Ah good point. I don't know why I did it that way. Fixed in [0eb7412](https://github.com/Layr-Labs/sidecar/pull/117/commits/0eb7412b60d8a3a7a25ed2d2316e3b5204aba3fd)

pkg/rewards/2_goldStakerRewardAmounts.go
```diff
@@ -107,26 +108,29 @@ staker_operator_total_tokens AS (
...
 token_breakdowns AS (
-  SELECT *,
+  SELECT sott.*,
     CASE
-      WHEN snapshot < @amazonHardforkDate AND reward_submission_date < @amazonHardforkDate THEN
-        cast(total_staker_operator_payout * 0.10 AS DECIMAL(38,0))
-      WHEN snapshot < @nileHardforkDate AND reward_submission_date < @nileHardforkDate THEN
-        (total_staker_operator_payout * 0.10)::text::decimal(38,0)
+      WHEN sott.snapshot < @amazonHardforkDate AND sott.reward_submission_date < @amazonHardforkDate THEN
+        cast(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / 10000.0 AS DECIMAL(38,0))
```
seanmcgary: I would change this in one of two ways:

1. Remove the decimal point from `10000.0` and let postgres have its way with the type casting
2. Explicitly cast `10000` to a known floating point type (probably `decimal`)

when I run `select pg_typeof(10000.0)` postgres treats it as numeric, which is likely what we want, but we should probably make that explicit for the sake of consistency (decimal is just an alias to numeric)

seanmcgary: Actually, we shouldnt need to touch this line at all. Same with the nile hardfork. None of these calculations are relevant any more and occurred in the past pre rewards v2, so we should leave them alone to prevent a regression.

cc: @ypatil12

pkg/rewards/operatorAvsSplitSnapshots.go
```diff
@@ -0,0 +1,92 @@
...
+	select
+		oas.operator,
+		oas.avs,
+		oas.activated_at::timestamp(6) as activated_at,
+		oas.new_operator_avs_split_bips as split,
+		oas.block_number,
+		oas.log_index,
+		b.block_time::timestamp(6) as block_time
+	from operator_avs_splits as oas
+	left join blocks as b on (b.number = oas.block_number)
```
seanmcgary: I know this is my fault for doing it elsewhere, but this should probably be just a `join` so that we require the block record to exist. Functionally, it probably doesnt change much, but doing a left join allows the block info to potentially be null

pkg/rewards/operatorPISplitSnapshots.go
```diff
@@ -0,0 +1,89 @@
...
+WITH operator_pi_splits_with_block_info as (
+	select
+		ops.operator,
+		ops.activated_at::timestamp(6) as activated_at,
+		ops.new_operator_avs_split_bips as split,
+		ops.block_number,
+		ops.log_index,
+		b.block_time::timestamp(6) as block_time
+	from operator_pi_splits as ops
+	left join blocks as b on (b.number = ops.block_number)
```
seanmcgary: Same as the comment above; should probably make this an inner join

seanmcgary (changes requested):
overall looks good. main comment that needs addressing is on the floating point math around token calculations and if we should even modify the logic for past forks since they're only ever used when backfilling

pkg/rewards/2_goldStakerRewardAmounts.go
ypatil12: Yup, agreed - much simpler!

pkg/rewards/2_goldStakerRewardAmounts.go
ypatil12: `floor(total_staker_operator_payout * 0.10)` is `numeric * numeric` since Postgres treats constants as numerics: 
 
> A numeric constant that contains neither a decimal point nor an exponent is initially presumed to be type integer if its value fits in type integer (32 bits); otherwise it is presumed to be type bigint if its value fits in type bigint (64 bits); otherwise it is taken to be type numeric. Constants that contain decimal points and/or exponents are always initially presumed to be type numeric.

I just want to be really sure that there is no float conversion. `numeric/numeric` should be numeric from my understanding but want a sanity check

pkg/rewards/operatorPISplitSnapshots.go
0xrajath: Done in [64bf833](https://github.com/Layr-Labs/sidecar/pull/117/commits/64bf8334c40fcf4505f74672a8341a96cd9858a2)

pkg/rewards/operatorAvsSplitSnapshots.go
0xrajath: Done in [64bf833](https://github.com/Layr-Labs/sidecar/pull/117/commits/64bf8334c40fcf4505f74672a8341a96cd9858a2)

pkg/rewards/2_goldStakerRewardAmounts.go
0xrajath: Done in [5c7f58c](https://github.com/Layr-Labs/sidecar/pull/117/commits/5c7f58cda135e3567ba668c9a5c152d780d59e4a)

pkg/rewards/2_goldStakerRewardAmounts.go
0xrajath: Done in [f5ce3ef](https://github.com/Layr-Labs/sidecar/pull/117/commits/f5ce3ef0a51a7b482d2b9a0c74e377ed7c21f06b)

pkg/rewards/2_goldStakerRewardAmounts.go
```diff
@@ -107,26 +108,29 @@ staker_operator_total_tokens AS (
...
-        cast(total_staker_operator_payout * 0.10 AS DECIMAL(38,0))
-      WHEN snapshot < @nileHardforkDate AND reward_submission_date < @nileHardforkDate THEN
-        (total_staker_operator_payout * 0.10)::text::decimal(38,0)
+      WHEN sott.snapshot < @amazonHardforkDate AND sott.reward_submission_date < @amazonHardforkDate THEN
+        cast(sott.total_staker_operator_payout * 0.10 AS DECIMAL(38,0))
+      WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
+        (sott.total_staker_operator_payout * 0.10)::text::decimal(38,0)
       ELSE
-        floor(total_staker_operator_payout * 0.10)
+        floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
```
ypatil12, line 123: I don't think this cast is actually doing anything? Since we already know 10000 will be a decimal by declaring it as a constant. My previous comment was moreso just sanity checking the types more than anything else

ypatil12: Oh, do you mean to cast the output of `split/10000` ?

0xrajath: This casting ensures that the resulting division will be of type `DECIMAL`. 

1. When `oas.split` is `NULL`:

`COALESCE(oas.split, 1000)` returns `1000` (an `INTEGER`).
`1000 / CAST(10000 AS DECIMAL)` results in `0.10` as `DECIMAL`, maintaining the same behavior as `0.10` prior to the change.

2. When `oas.split` is not `NULL`:

The result will still be computed with `DECIMAL` precision, ensuring no loss of precision during the division.

0xrajath: Also suggested by @seanmcgary here: https://github.com/Layr-Labs/sidecar/pull/117#discussion_r1856795893

0xrajath: i.e `INTEGER` / `DECIMAL` in Postgres ensures that the resulting value is of type `DECIMAL`.

0xrajath: Also the reason to do this way vs cast `split/10000` is so that precision is maintained. In the former we multiply by numerator first and then divde by denominator. Vs for the latter we do division first leading to precision loss (albeit not too much).

ypatil12: Okay, all for the explicitness here

ypatil12 (dismissed):

pkg/rewards/2_goldStakerRewardAmounts.go
```diff
@@ -107,26 +108,29 @@ staker_operator_total_tokens AS (
...
-        total_staker_operator_payout - cast(total_staker_operator_payout * 0.10 as DECIMAL(38,0))
-      WHEN snapshot < @nileHardforkDate AND reward_submission_date < @nileHardforkDate THEN
-        total_staker_operator_payout - ((total_staker_operator_payout * 0.10)::text::decimal(38,0))
+      WHEN sott.snapshot < @amazonHardforkDate AND sott.reward_submission_date < @amazonHardforkDate THEN
+        sott.total_staker_operator_payout - cast(sott.total_staker_operator_payout * 0.10 as DECIMAL(38,0))
+      WHEN sott.snapshot < @nileHardforkDate AND sott.reward_submission_date < @nileHardforkDate THEN
+        sott.total_staker_operator_payout - ((sott.total_staker_operator_payout * 0.10)::text::decimal(38,0))
       ELSE
-        total_staker_operator_payout - floor(total_staker_operator_payout * 0.10)
+        sott.total_staker_operator_payout - floor(sott.total_staker_operator_payout * COALESCE(oas.split, 1000) / CAST(10000 AS DECIMAL))
```
seanmcgary, line 133: @ypatil12 what are your thoughts on adding a hardfork here rather than modifying in place as a general practice?

My concern here (having dealt with Postgres' finicky precision) is that even though this _shouldnt_ change anything in the past, Im not 100% sure it wont. Adding a hardfork guarantees that it wont

0xrajath: Going with the hardfork approach.

0xrajath: Done in [5095472](https://github.com/Layr-Labs/sidecar/pull/117/commits/50954729794dadc3fe6682128513821e1d1f5679)

seanmcgary (approved):

PR #143, opened 2024-12-08
feat: bcsf slashing

gpsanant:
Adds accounting for slashing from the beacon chain


seanmcgary:
Closed and re-opened as #158

PR #146, opened 2024-12-11
feat: add create-snapshot and restore-snapshot commands to the cli

BrendanCoughlan5:
Introduces two new cli commands: `create-snapshot` and `restore-snapshot` to allow users to easily backup their sidecar databases and bootstrap faster by restoring from an existing snapshot.


cmd/snapshot-create.go
seanmcgary, line 1: Should rename to `createSnapshot.go`. Files should always be camelCase and when creating a command, it should be phrased in a command as if you were saying it aloud.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+
+var (
+	snapshotCreateToPath string
+	snapshotFidelity     string
+	s3Path               string
+	compressSnapshot     bool
+)
+
+var snapshotCreateCmd = &cobra.Command{
+	Use:   "snapshot-create",
```
seanmcgary: `"create-snapshot"`

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+	Short: "Create a snapshot of the database",
+	Long: `Create a snapshot of the database.
+
+Currently available fidelity levels:
+- archive (default): includes chain data, EigenModel state, rewards, and staker-operator table data.
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Initialize logger (in a real scenario, you'd likely pass in a config)
+		l, _ := logger.NewLogger(&logger.LoggerConfig{Debug: true})
```
seanmcgary: Look at this file for inspiration on how to set things up properly:

https://github.com/Layr-Labs/sidecar/blob/master/cmd/database.go#L35-L109

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	"github.com/spf13/cobra"
+)
+
+var (
+	snapshotCreateToPath string
+	snapshotFidelity     string
```
seanmcgary: Rename to `snapshotType` and make it an enum to show the different types available.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	"github.com/spf13/cobra"
+)
+
+var (
+	snapshotCreateToPath string
```
seanmcgary: Rename to `output` that specifically takes a file path and also consider defaulting to writing to stdout if `--output` is omitted.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+	snapshotFidelity     string
+	s3Path               string
```
seanmcgary: Going back to my comment above about allowing output on stdout, if you do that, you can reduce the internal complexity of the sidecar to not know about S3 at all.

```bash
sidecar generate-snapshot | aws s3 cp - s3://<bucket name>/<path>/<file>.dump
```

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+	compressSnapshot     bool
```
seanmcgary: This should likely be more specific, or offer a flag to the tune of `--compression-type=<gzip, zip, etc>`. Omitting leaves it uncompressed.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+		dbUser, _ := cmd.Flags().GetString("database.user")
+		dbPassword, _ := cmd.Flags().GetString("database.password")
+		dbPort, _ := cmd.Flags().GetInt("database.port")
+		schemaName, _ := cmd.Flags().GetString("database.schema_name")
+
+		// Retrieve chain and fidelity
+		chain, _ := cmd.Flags().GetString("chain")
+
+		// Validate fidelity
+		if snapshotFidelity != "archive" {
```
seanmcgary: Good to check invariants, but a default should be set via cobra for the corresponding flag.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+		if snapshotCreateToPath == "" {
+			fileNameSuffix := time.Now().UTC().Format("2006-01-02-15-04-05")
+			snapshotCreateToPath = fmt.Sprintf("sidecar-snapshot-%s-%s-%s.dump", chain, snapshotFidelity, fileNameSuffix)
+		}
+
+		// Log the snapshot file path
+		l.Sugar().Infow("Snapshot will be created at this file", "file", snapshotCreateToPath)
+
+		// Set the password in an environment variable for pg_dump
+		if err := os.Setenv("PGPASSWORD", dbPassword); err != nil {
```
seanmcgary: This should be set as a connection string not as an environment variable.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+			l.Sugar().Fatalw("Failed to set environment variable", "error", err)
+			return err
+		}
+
+		// Construct the pg_dump command with a context for timeout
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
seanmcgary: Default should likely be no timeout at all. Some of these snapshots may take a long time and the worst thing is bumping into a timeout when your dump was almost done and have to retry and wait even longer.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+		defer cancel()
+
+		pgDumpCmd := exec.CommandContext(ctx,
```
seanmcgary: We should only shell out as a last resort. [habx/pg-commands](https://github.com/habx/pg-commands) looks like a promising library that can simplify this process.

cmd/snapshot-create.go
```diff
@@ -0,0 +1,161 @@
...
+				"error", err,
+				"command", pgDumpCmd.String(),
+			)
+			return fmt.Errorf("failed to run pg_dump: %w", err)
+		}
+
+		l.Sugar().Infow("Successfully created snapshot", "file", snapshotCreateToPath)
+
+		// Compress the snapshot if the flag is set
+		if compressSnapshot {
```
seanmcgary: Rather than downloading the file to the filesystem and THEN compressing it, Id suggest looking into how to pipeline the process to minimize how much we write then read to disk. If you can take the stream from `pg_dump` and optionally pipe it directly to `tar`, thats much preferred than writing to disk first THEN compressing

seanmcgary (changes requested):
Start with these comments, most will apply to the restore command as well.

README.md
```diff
@@ -246,15 +246,40 @@ POSTGRES_DATA_PATH=<path to store postgres data> docker-compose up
 
 # Boot from a snapshot
 
+Get our latest snapshots
 * Mainnet (not yet available)
-* Testnet ([2024-11-22](https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky-20241122.tar.gz))
+* Testnet 
+    * Latest archive snapshot (~1.5GB) https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/holesky/archive/sidecar-snapshot-holesky-archive-latest.tar.gz
```
seanmcgary: Two things:

* Snapshots should have dates in them
* We should say what "latest" means, especially since we're not running this regularly yet

I would also double check your size. The last testnet snapshot I took on Monday was 3.23GB, so more than double the 1.5GB you have listed.

README.md
```diff
@@ -246,15 +246,40 @@ POSTGRES_DATA_PATH=<path to store postgres data> docker-compose up
...
 
+
+Example script
 ```bash
-curl -LO https://eigenlayer-sidecar.s3.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky-20241122.tar.gz
+curl -LO https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/holesky/archive/sidecar-snapshot-holesky-archive-latest.tar.gz
+
+tar -xvf sidecar-snapshot-holesky-archive-latest.tar.gz
+
+go run main.go snapshot-restore \
```
seanmcgary: `snapshot-restore` should be able to take the URL to the file directly and use it. You shouldnt have to download and unpack it in a separate step.

README.md
```diff
@@ -246,15 +246,40 @@ POSTGRES_DATA_PATH=<path to store postgres data> docker-compose up
...
+  --clean=true \
+  --database.host=localhost \
+  --database.user=sidecar \
+  --database.password=... \
+  --database.port=5432 \
+  --database.db_name=sidecar
+```
 
-tar -xvf sidecar-testnet-2024-11-22.tar.gz
+Browse our public snapshots here
```
seanmcgary: No need for this yet, two reasons:

1. The bucket isnt browse-able directly for security reasons
2. We'll want to eventually create a dedicated page listing snapshots we've created.

What you have at the top of the readme is fine for now

cmd/createSnapshot.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		// Validate Type
+		if snapshotType != "archive" {
+			l.Sugar().Warnw("Unsupported Type specified; falling back to 'archive'.",
+				"requested", snapshotType,
+				"used", "archive",
+			)
+			snapshotType = "archive"
+		}
```
seanmcgary: Flags are still implemented incorrectly. All of these values should be referenced from the instance of `config.Config` like the other CLI commands. 

https://github.com/Layr-Labs/sidecar/blob/master/cmd/run.go#L41

Let me know if you need help with this

cmd/createSnapshot.go
```diff
@@ -0,0 +1,160 @@
...
+
+		// Execute pg_dump command
+		cmdExec := exec.Command(dumpCmd[0], dumpCmd[1:]...)
+		outputBytes, err := cmdExec.CombinedOutput()
+		if err != nil {
+			l.Sugar().Fatalw("Failed to create database snapshot", "error", err, "output", string(outputBytes))
+			os.Exit(1)
+		}
+
+		l.Sugar().Infow("Successfully created snapshot", "file", output)
```
seanmcgary: Dumping the likely multi-GB file to a log seems like a bad idea

cmd/createSnapshot.go
```diff
@@ -0,0 +1,160 @@
...
+		if schemaName != "" {
+			dumpCmd = append(dumpCmd, "-n", schemaName)
+		}
+
+		if output != "" {
+			// If output is specified, we consider output as a full file path
+			dumpDir := filepath.Dir(output)
+
+			// Ensure the dump directory exists
+			if err := os.MkdirAll(dumpDir, 0755); err != nil {
```
seanmcgary: Standard cli practice is to _not_ assume that the user wants the directory created unless explicitly noted somewhere. If it doesnt exists, I would error out and let the caller create it with their desired permissions level.

cmd/createSnapshot.go
```diff
@@ -0,0 +1,160 @@
...
+				l.Sugar().Fatalw("Failed to create output directory", "directory", dumpDir, "error", err)
+				os.Exit(1)
```
seanmcgary: `l.Sugar().Fatal()` will terminate the process automatically. If you want to provide a custom exit code, this should be `l.Sugar().Error`.

cmd/createSnapshot.go
```diff
@@ -0,0 +1,160 @@
...
+		outputBytes, err := cmdExec.CombinedOutput()
+		if err != nil {
+			l.Sugar().Fatalw("Failed to create database snapshot", "error", err, "output", string(outputBytes))
+			os.Exit(1)
+		}
+
+		l.Sugar().Infow("Successfully created snapshot", "file", output)
+
+		// Upload to S3 if s3Path is specified
+		if s3Path != "" {
```
seanmcgary: We should not take a dependency on S3 in the sidecar

seanmcgary (approved):

README.md
```diff
@@ -250,12 +252,112 @@ POSTGRES_DATA_PATH=<path to store postgres data> docker-compose up
...
 
+
+## Example boot from testnet snapshot
 ```bash
 curl -LO https://eigenlayer-sidecar.s3.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky-20241122.tar.gz
 
 tar -xvf sidecar-testnet-2024-11-22.tar.gz
 
-pg_restore --host <hostname> --port 5432 --username <username> --dbname <dbname> --no-owner sidecar-testnet-2024-11-22.dump
+go run main.go restore-snapshot \
```
seanmcgary: rather than referencing `go run` this should reference the sidecar binary

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+
+var createSnapshotCmd = &cobra.Command{
+	Use:   "create-snapshot",
+	Short: "Create a snapshot of the database",
+	Long:  "Create a snapshot of the database.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		initCreateSnapshotCmd(cmd)
+		cfg := config.NewConfig()
+
+		// Initialize logger
```
seanmcgary: No need for this comment

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+		l, err := logger.NewLogger(&logger.LoggerConfig{Debug: cfg.Debug})
+		if err != nil {
+			return fmt.Errorf("failed to initialize logger: %w", err)
+		}
+		defer func() {
+			if syncErr := l.Sync(); syncErr != nil {
```
seanmcgary: Not needed since zap loggers are unbuffered by deafult

internal/config/config.go
```diff
@@ -333,3 +334,25 @@ func StringVarToInt(val string) int {
...
+	if dbConfig.Host == "" {
+		return fmt.Errorf("database host is missing")
+	}
+	if dbConfig.DbName == "" {
+		return fmt.Errorf("database name is missing")
+	}
+	if dbConfig.User == "" {
+		return fmt.Errorf("database user is missing")
+	}
+	if dbConfig.Password == "" {
```
seanmcgary: An empty password is actually technically valid

internal/config/config.go
```diff
@@ -333,3 +334,25 @@ func StringVarToInt(val string) int {
...
+		return fmt.Errorf("database password is missing")
+	}
+	if dbConfig.Port == 0 {
```
seanmcgary: This should default to 5432 automatically if no port is provided

internal/config/config.go
```diff
@@ -333,3 +334,25 @@ func StringVarToInt(val string) int {
...
+		return fmt.Errorf("database port is missing or invalid")
+	}
+	if dbConfig.SchemaName == "" {
```
seanmcgary: This is fine as an empty string and defaults to `public` automatically

internal/config/config.go
```diff
@@ -333,3 +334,25 @@ func StringVarToInt(val string) int {
...
 }
+
+func ValidateDatabaseConfig(dbConfig DatabaseConfig) error {
+	if dbConfig.Host == "" {
+		return fmt.Errorf("database host is missing")
+	}
+	if dbConfig.DbName == "" {
+		return fmt.Errorf("database name is missing")
+	}
+	if dbConfig.User == "" {
```
seanmcgary: Technically not necessary; postgres will assume the username of `whoami` running the process.

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+				l.Sugar().Errorw("Failed to sync logger", "error", syncErr)
+			}
+		}()
+
+		// Validate database configuration
+		dbConfig := cfg.DatabaseConfig
+		if err := config.ValidateDatabaseConfig(dbConfig); err != nil {
```
seanmcgary: Config validation should ideally happen as part of the postgres module

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+			l.Sugar().Fatalw("Invalid database configuration", "error", err)
+		}
+
+		// Log database connection details without password
+		l.Sugar().Infow("Database connection details",
```
seanmcgary: Not necessary

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	pgcommands "github.com/habx/pg-commands" // Imported as pgcommands
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+var (
+	output string
```
seanmcgary: This shouldnt be a global variable

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+			fmt.Printf("Failed to bind flag '%s' - %+v\n", f.Name, err)
+		}
+		if err := viper.BindEnv(f.Name); err != nil {
+			fmt.Printf("Failed to bind env '%s' - %+v\n", f.Name, err)
+		}
+	})
+}
+
+func init() {
+	createSnapshotCmd.Flags().StringVarP(&output, "output", "f", "", "Path to save the snapshot file to (required)")
```
seanmcgary: This should follow the other commands and should be part of the database config object, assembled in the config module by reading from viper.

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+		if err := viper.BindPFlag(config.KebabToSnakeCase(f.Name), f); err != nil {
+			fmt.Printf("Failed to bind flag '%s' - %+v\n", f.Name, err)
+		}
+		if err := viper.BindEnv(f.Name); err != nil {
+			fmt.Printf("Failed to bind env '%s' - %+v\n", f.Name, err)
+		}
+	})
+}
+
+func init() {
```
seanmcgary: I dont see this init function called anywhere

cmd/restoreSnapshot.go
```diff
@@ -0,0 +1,117 @@
...
+		if err := viper.BindPFlag(config.KebabToSnakeCase(f.Name), f); err != nil {
+			fmt.Printf("Failed to bind flag '%s' - %+v\n", f.Name, err)
+		}
+		if err := viper.BindEnv(f.Name); err != nil {
+			fmt.Printf("Failed to bind env '%s' - %+v\n", f.Name, err)
+		}
+	})
+}
+
+func init() {
```
seanmcgary: Same - dont see this called anywhere and this is technically a naming collision since restoreSnapshot.go and createSnapshot.go are in the same module

cmd/restoreSnapshot.go
```diff
@@ -0,0 +1,117 @@
...
+	restoreSnapshotCmd.Flags().StringVarP(&restoreSnapshotFromPath, "restore-snapshot-from-path", "f", "", "Path to the snapshot file (required)")
+	if err := restoreSnapshotCmd.MarkFlagRequired("restore-snapshot-from-path"); err != nil {
```
seanmcgary: `snapshot-file` is a better name

internal/config/config.go
```diff
@@ -333,3 +334,25 @@ func StringVarToInt(val string) int {
 	}
 	return v
 }
+
+func ValidateDatabaseConfig(dbConfig DatabaseConfig) error {
+	if dbConfig.Host == "" {
```
seanmcgary: Host is typically optionally and localhost is assumed if omitted

seanmcgary (changes requested):

cmd/createSnapshot.go
```diff
@@ -0,0 +1,114 @@
...
+			l.Sugar().Fatal("Output path must be specified")
+		}
+
+		// Execute the dump command
+		dumpExec := dump.Exec(pgcommands.ExecOptions{StreamPrint: false})
+		if dumpExec.Error != nil {
+			l.Sugar().Fatalw("Failed to create database snapshot", "error", dumpExec.Error.Err, "output", dumpExec.Output)
+		}
+
+		l.Sugar().Infow("Successfully created snapshot")
```
seanmcgary: Let's move this logic into a module rather than having it all live in-line in the cmd directly. While doing that, tests should also be added

cmd/restoreSnapshot.go
```diff
@@ -0,0 +1,117 @@
...
+		// Execute the restore
+		restoreExec := restore.Exec(restoreSnapshotFromPath, pgcommands.ExecOptions{StreamPrint: false})
+		if restoreExec.Error != nil {
+			l.Sugar().Fatalw("Failed to restore from snapshot",
+				"error", restoreExec.Error.Err,
+				"output", restoreExec.Output,
+			)
+		}
+
+		l.Sugar().Infow("Successfully restored from snapshot")
```
seanmcgary: Same here - this should live in a module with tests

internal/config/config.go
```diff
@@ -46,12 +47,14 @@ type EthereumRpcConfig struct {
...
-	Password   string
-	DbName     string
-	SchemaName string
+	Host                 string
+	Port                 int
+	User                 string
+	Password             string
+	DbName               string
+	SchemaName           string
+	CreateSnapshotOutput string
```
seanmcgary: Doesnt belong on the DatabaseConfig object and should be in a separate config for creating snapshots or restoring from a snapshot

internal/config/config.go
```diff
@@ -46,12 +47,14 @@ type EthereumRpcConfig struct {
...
+	RestoreSnapshotFile  string
```
seanmcgary: Same here - see above comment

cmd/root.go
```diff
@@ -39,6 +46,8 @@ func init() {
 	rootCmd.PersistentFlags().String(config.DatabasePassword, "", `PostgreSQL password`)
 	rootCmd.PersistentFlags().String(config.DatabaseDbName, "sidecar", `PostgreSQL database name`)
 	rootCmd.PersistentFlags().String(config.DatabaseSchemaName, "", `PostgreSQL schema name (default "public")`)
+	createSnapshotCmd.PersistentFlags().String(config.DatabaseCreateSnapshotOutput, "", "Path to save the snapshot file to (required)")
```
seanmcgary: Since both of these shouldnt be scoped to the database config namespace, let's move them to the bottom, just before 

```
rootCmd.PersistentFlags().VisitAll(func(f *pflag.Flag) {
```

seanmcgary (changes requested):

internal/config/config.go
seanmcgary: Also this variable name is confusing. As worded I would think it's a boolean but it isnt. `--output` or `--output-file` are both much better and are what I recommended previously https://github.com/Layr-Labs/sidecar/pull/146#discussion_r1881251277

internal/config/config.go
seanmcgary: Same here - variable sounds like a boolean but isnt. `--input` or `--input-file` would be better, especially since this value is scoped only to the restore-snapshot command

cmd/createSnapshot.go
```diff
@@ -0,0 +1,95 @@
...
+			dump.EnableVerbose()
+		}
+
+		// Set schema
+		if dbConfig.SchemaName != "" {
+			dump.Options = append(dump.Options, fmt.Sprintf("--schema=%s", dbConfig.SchemaName))
+		}
+
+		// Set up output
+		outputPath := dbConfig.CreateSnapshotOutput
```
seanmcgary: Is this expected to be a path or the destination file?

seanmcgary (changes requested):

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize logger: %w", err)
+	}
+	return l, nil
+}
+
+func validateSnapshotConfig(cfg *config.Config, l *zap.Logger) (*config.DatabaseConfig, error) {
+	dbConfig := cfg.DatabaseConfig
+
+	l.Sugar().Debugw("Database connection details",
```
seanmcgary: Per my previous comment, this isnt needed

https://github.com/Layr-Labs/sidecar/pull/146#discussion_r1902013868

pkg/snapshot/snapshot.go
seanmcgary, line 1: rather than creating a bunch of functions that take the same arguments over and over (config and logger), this functionality should be encapsulated in a proper struct that is initialized with said config and logger

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	pgcommands "github.com/habx/pg-commands"
+	"go.uber.org/zap"
+)
+
+// CreateSnapshot creates a snapshot of the database based on the provided configuration.
+func CreateSnapshot(cfg *config.Config) error {
+	l, err := initializeLogger(cfg)
```
seanmcgary: Its not the responsibility of this file to create the logger instance; the caller should be responsible for that so that the logger instance isnt coupled to this function

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+
+	return executeSnapshotDump(dump, l)
+}
+
+// RestoreSnapshot restores a snapshot of the database based on the provided configuration.
+func RestoreSnapshot(cfg *config.Config) error {
+	l, err := initializeLogger(cfg)
```
seanmcgary: Same here - creation of the logger shouldnt be coupled to this module

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+		"user", dbConfig.User,
+		"port", dbConfig.Port,
+		"schema", dbConfig.SchemaName,
+	)
+
+	if cfg.SnapshotConfig.OutputFile == "" {
+		return nil, fmt.Errorf("output path i.e. `snapshot-output-file` must be specified")
+	}
+
+	return &dbConfig, nil
```
seanmcgary: This is an anti-pattern that can introduce unwanted side-effects. if you're going to create a function to validate the contents of the config, it should return a boolean indicating if it is valid or not. alternatively an error if you want to return more information that a simple true/false

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func setupSnapshotDump(cfg *config.Config, dbConfig *config.DatabaseConfig, l *zap.Logger) (*pgcommands.Dump, error) {
+	if dbConfig.Host == "" {
```
seanmcgary: Why isnt this check in the `validateSnapshotConfig` function?

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if dbConfig.SchemaName != "" {
+		dump.Options = append(dump.Options, fmt.Sprintf("--schema=%s", dbConfig.SchemaName))
+	}
+
+	if cfg.Debug {
+		dump.EnableVerbose()
+	}
+
+	outputPath := cfg.SnapshotConfig.OutputFile
```
seanmcgary: no need for the intermediate variable, just use it directly

```go
if err := validateOutputPath(cfg.SnapshotConfig.OutputFile, l); err != nil {
```

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+	if err := validateOutputPath(outputPath, l); err != nil {
+		return nil, err
+	}
+	dump.SetFileName(outputPath)
+
+	return dump, nil
+}
+
+func executeSnapshotDump(dump *pgcommands.Dump, l *zap.Logger) error {
```
seanmcgary: this function serves little purpose over simply calling 

```go
dumpExec := dump.Exec(pgcommands.ExecOptions{StreamPrint: false})
if dumpExec.Error != nil {
    l.Sugar().Fatalw("Failed to create database snapshot", "error", dumpExec.Error.Err, "output", dumpExec.Output)
    return dumpExec.Error.Err
}
```

directly instead

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+		Username: dbConfig.User,
+		Password: dbConfig.Password,
+	})
+	if err != nil {
+		l.Sugar().Fatalw("Failed to initialize restore", "error", err)
+		return nil, err
+	}
+
+	restore.ResetOptions()
+	restore.Options = []string{"--no-owner", "--no-acl", "--exit-on-error"}
```
seanmcgary: Not sure why these are here since they are the defaults provided in the library

https://github.com/habx/pg-commands/blob/dev/pg_restore.go#L13

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,198 @@
...
+		restore.EnableVerbose()
+	}
+	if dbConfig.SchemaName != "" {
+		restore.SetSchemas([]string{dbConfig.SchemaName})
+	}
+
+	return restore, nil
+}
+
+func executeRestore(restore *pgcommands.Restore, snapshotFile string, l *zap.Logger) error {
```
seanmcgary: Similar to `executeSnapshotDump` above, this function serves little purpose. better to call 

```go
restoreExec := restore.Exec(snapshotFile, pgcommands.ExecOptions{StreamPrint: false})
if restoreExec.Error != nil {
    l.Sugar().Fatalw("Failed to restore from snapshot",
	"error", restoreExec.Error.Err,
	"output", restoreExec.Output,
    )
    return restoreExec.Error.Err
}
```

directly than wrap it in a function

pkg/snapshot/snapshot_test.go
seanmcgary, line 1: - `RestoreSnapshot` needs coverage
- `CreateSnapshot` needs coverage

internal/config/config.go
```diff
@@ -104,6 +111,9 @@ var (
 	DatabaseDbName     = "database.db_name"
 	DatabaseSchemaName = "database.schema_name"
 
+	SnapshotOutputFile = "snapshot.output_file"
+	SnapshotInputFile  = "snapshot.input_file"
```
seanmcgary: Like I mentioned before, since these flags are scoped to `create-snapshot` and `restore-snapshot`, they dont need a `snapshot.` prefix.

https://github.com/Layr-Labs/sidecar/pull/146#discussion_r1904716948

seanmcgary (changes requested):

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	pgcommands "github.com/habx/pg-commands"
+	"go.uber.org/zap"
+)
+
+// SnapshotConfig encapsulates the configuration and logger for snapshot operations.
+type SnapshotConfig struct {
+	cfg *config.Config
+	l   *zap.Logger
```
seanmcgary, line 28: configs should be data only and not include instances of a other modules.

for this `SnapshotConfig` to be meaningful, it shouldnt just wrap the top level app config; it should specify exactly what it needs. In this case, that is like the database credentials/details and input/output file names. It shouldnt be directly coupled to the global config instance.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// NewSnapshotConfig initializes a new SnapshotConfig with the given configuration and logger.
+func NewSnapshotConfig(cfg *config.Config, l *zap.Logger) *SnapshotConfig {
```
seanmcgary: Related to my above comment about the config, to get proper encapsulation of the snapshot config and logger instance, you should really have a `type SnapshotService struct` type that stores a config and the logger and then scope the below functions to said struct.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+		Password: dbConfig.Password,
+	})
+	if err != nil {
+		sCfg.l.Sugar().Errorw("Failed to initialize pg-commands Dump", "error", err)
+		return nil, err
+	}
+
+	// Not using the default pg-commands options to avoid using --clean
+	dump.ResetOptions()
+	dump.Options = append(dump.Options, "--no-owner", "--no-acl", "--blob")
```
seanmcgary: I would say we probably WANT to use `--clean`. If someone tries to apply a snapshot to a database that already has data in it, the act of restoring should drop any existing data and apply the snapshot. Otherwise, you'll end up with corrupt data

seanmcgary: Would also make sure this is using `-Fc` as the output type if its not already doing so.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil || info.IsDir() {
+		sCfg.l.Sugar().Errorw("Snapshot file does not exist", "path", snapshotFile)
+		return false
+	}
+
+	return true
+}
+
+func setupRestore(sCfg *SnapshotConfig) (*pgcommands.Restore, error) {
+	dbConfig := &sCfg.cfg.DatabaseConfig
```
seanmcgary: No need for the intermediate assignment, just reference the values directly

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+		)
+		return restoreExec.Error.Err
+	}
+
+	sCfg.l.Sugar().Infow("Successfully restored from snapshot")
+	return nil
+}
+
+func validateCreateSnapshotConfig(sCfg *SnapshotConfig) bool {
+	dbConfig := sCfg.cfg.DatabaseConfig
```
seanmcgary: No need for the intermediate assignment, just reference the values directly

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if err := validateOutputPath(sCfg); err != nil {
+		return false
+	}
+
+	return true
+}
+
+func setupSnapshotDump(sCfg *SnapshotConfig) (*pgcommands.Dump, error) {
+	dbConfig := sCfg.cfg.DatabaseConfig
```
seanmcgary: No need for the intermediate assignment, just reference the values directly

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+
+	dump.SetFileName(sCfg.cfg.SnapshotConfig.OutputFile)
+
+	return dump, nil
+}
+
+func validateRestoreConfig(sCfg *SnapshotConfig) bool {
+	snapshotFile := sCfg.cfg.SnapshotConfig.InputFile
+	if snapshotFile == "" {
+		sCfg.l.Sugar().Error("Restore snapshot file path i.e. `snapshot-input-file` must be specified")
```
seanmcgary: `input-file`

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if dbConfig.SchemaName != "" {
+		restore.SetSchemas([]string{dbConfig.SchemaName})
+	}
+
+	return restore, nil
+}
+
+func validateOutputPath(sCfg *SnapshotConfig) error {
+	outputPath := sCfg.cfg.SnapshotConfig.OutputFile
```
seanmcgary: No need for the intermediate assignment, just reference the values directly

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,176 @@
...
+	outputDir := filepath.Dir(outputPath)
+	if _, err := os.Stat(outputDir); os.IsNotExist(err) {
+		sCfg.l.Sugar().Errorw("Snapshot output directory does not exist", "directory", outputDir)
```
seanmcgary: This is slightly confusing because we're providing a file, but then logging output about its directory...would it be better to let this error come from pg_dump directly if it fails to write to the provided file name?

seanmcgary (changes requested):

pkg/postgres/postgres.go
```diff
@@ -28,7 +29,7 @@ type Postgres struct {
 	Db *sql.DB
 }
 
-func GetTestPostgresDatabase(cfg config.DatabaseConfig, gCfg *config.Config, l *zap.Logger) (
+func GetTestPostgresDatabase(cfg config.DatabaseConfig, gCfg *config.Config, l *zap.Logger, noMigrations ...bool) (
```
seanmcgary: why is this unpacking an array and not just a simple bool?

pkg/postgres/postgres.go
```diff
@@ -53,9 +54,12 @@ func GetTestPostgresDatabase(cfg config.DatabaseConfig, gCfg *config.Config, l *
 		return testDbName, nil, nil, err
 	}
 
-	migrator := migrations.NewMigrator(pg.Db, grm, l, gCfg)
-	if err = migrator.MigrateAll(); err != nil {
-		return testDbName, nil, nil, err
+	// Check if noMigrations is set to true, default is false
+	if !(len(noMigrations) > 0 && noMigrations[0]) {
```
seanmcgary: Im confused as to why noMigrations is an array and not a bool

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		s.l.Sugar().Errorw("Failed to initialize pg-commands Dump", "error", err)
+		return nil, err
+	}
+
+	if s.dCfg.SchemaName != "" {
+		dump.Options = append(dump.Options, fmt.Sprintf("--schema=%s", s.dCfg.SchemaName))
+	}
+
+	if s.debug {
```
seanmcgary: Rather than `debug` this should more closely be mapped to what it actually means in this context, which is enabling verbose mode, so I would say this should be renamed to `verbose`. Debug in the broader context is intended for setting the log level specifically.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,180 @@
...
+	info, err := os.Stat(snapshotFile)
+	if err != nil || info.IsDir() {
+		s.l.Sugar().Errorw("Snapshot file does not exist", "path", snapshotFile)
+		return false
+	}
+
+	return true
+}
+
+func userCurrent() (string, error) {
```
seanmcgary: why are we doing this?

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,180 @@
...
+		Username: s.dCfg.User,
+		Password: s.dCfg.Password,
+	})
+	if err != nil {
+		s.l.Sugar().Fatalw("Failed to initialize restore", "error", err)
+		return nil, err
+	}
+
+	restore.Options = append(restore.Options, "--if-exists")
+	if s.dCfg.User == "" {
```
seanmcgary: if user is empty, then pg_restore and pg_dump will automatically determine which user to use so this isnt necessary.

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,180 @@
...
+		user, err := userCurrent()
+		if err != nil {
+			s.l.Sugar().Fatalw("Failed to determine current user", "error", err)
+			return nil, err
+		}
+		s.dCfg.User = user
+	}
+	restore.Role = s.dCfg.User
+
+	if s.debug {
```
seanmcgary: same as above, this should be `s.verbose`

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	pgcommands "github.com/habx/pg-commands"
+	"go.uber.org/zap"
+)
+
+// SnapshotService encapsulates the configuration and logger for snapshot operations.
+type SnapshotService struct {
+	dCfg  *config.DatabaseConfig
+	sCfg  *config.SnapshotConfig
```
seanmcgary: why do we have two configs? really we should only have one config that includes everything we need, otherwise you end up making the snapshot module take a dependency on other module configs, which it shouldnt.

From my previous [comment](https://github.com/Layr-Labs/sidecar/pull/146#discussion_r1907498515):

> for this SnapshotConfig to be meaningful, it shouldnt just wrap the top level app config; it should specify exactly what it needs. In this case, that is like the database credentials/details and input/output file names. It shouldnt be directly coupled to the global config instance.

e.g....

_config.go_
```go
type SnapshotConfig struct {
	OutputFile string
	InputFile  string
}
```

_snapshot.go_
```go
type SnapshotConfig struct {
    OutputFile string
    InputFile  string
    Host       string
    Port       int
    User       string
    Password   string
    DbName     string
    SchemaName string
}
```

pkg/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+	error,
+) {
+	cfg := config.NewConfig()
+	cfg.Chain = config.Chain_Mainnet
+	cfg.Debug = false
+	cfg.DatabaseConfig = *tests.GetDbConfigFromEnv()
+
+	l, _ := logger.NewLogger(&logger.LoggerConfig{Debug: true})
+
+	dbname, _, grm, err := postgres.GetTestPostgresDatabase(cfg.DatabaseConfig, cfg, l, noMigrations)
```
seanmcgary: still confused as to why `GetTestPostgresDatabase` expects a variable number of bools as the final arguments.

I would also recommend refactoring `GetTestPostgresDatabase` to be more composable rather than taking a boolean to drive behavior.

pkg/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func TestCreateAndRestoreSnapshot(t *testing.T) {
+	tempDir := t.TempDir()
+	dumpFile := filepath.Join(tempDir, "test_snapshot.dump")
+
+	dbName1, db1grm, l, db1cfg, db1err := setupDatabase(false)
+	if db1err != nil {
+		t.Fatal(db1err)
+	}
```
seanmcgary: this should be scoped to the first `t.Run()` block (which is kinda the whole point)

pkg/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	dbName2, db2grm, _, db2cfg, db2err := setupDatabase(true)
+	if db2err != nil {
+		t.Fatal(db2err)
+	}
```
seanmcgary: Same here - scoped to the second t.Run() block

pkg/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	t.Run("Create snapshot from db1", func(t *testing.T) {
+		db1cfg.DatabaseConfig.DbName = dbName1
+
+		svc := NewSnapshotService(&db1cfg.DatabaseConfig, &config.SnapshotConfig{OutputFile: dumpFile}, l, false)
+		err := svc.CreateSnapshot()
+		assert.NoError(t, err, "Creating snapshot should not fail")
```
seanmcgary, line 179: should probably also assert that the output file isnt empty

pkg/snapshot/snapshot_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+		var db2Count int
+		db2grm.Raw("select count(*) from migrations").Scan(&db2Count)
+
+		assert.Equal(t, db1Count, db2Count, "The number of migrations in db1 and db2 should be equal")
+	})
+	t.Cleanup(func() {
+		postgres.TeardownTestDatabase(dbName1, db1cfg, db1grm, l)
+		postgres.TeardownTestDatabase(dbName2, db2cfg, db2grm, l)
+		os.RemoveAll(tempDir)
```
seanmcgary: I wouldnt nuke the entire tmp directory since you dont know what else is using that. Just remove the single file you create.

seanmcgary (changes requested):

pkg/snapshot/snapshot_test.go
seanmcgary: `os.Stat` isnt enough here. You need to check the the `.Size` parameter and probably ensure its larger than 4k since even empty files technically take up space on disk

pkg/postgres/postgres.go
```diff
@@ -33,6 +34,25 @@ func GetTestPostgresDatabase(cfg config.DatabaseConfig, gCfg *config.Config, l *
...
+
+	migrator := migrations.NewMigrator(pg, grm, l, gCfg)
+	if err = migrator.MigrateAll(); err != nil {
+		return testDbName, nil, nil, err
+	}
+
+	return testDbName, pg, grm, nil
+}
+
+func GetTestPostgresDatabaseNoMigrations(cfg config.DatabaseConfig, l *zap.Logger) (
```
seanmcgary: By more composable I meant:

`GetTestPostgresDatabase()` <-- get the database, but dont run the migrator
`GetTestPostgresDatabaseWithMigrations()` <-- should call `GetTestPostgresDatabase` and subsequently run the migrations.

But to do this without breaking things, youd first need to rename `GetTestPostgresDatabase` across the project to `GetTestPostgresDatabaseWithMigrations` then re-introduce `GetTestPostgresDatabase`

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	dump, err := s.setupSnapshotDump()
+	if err != nil {
+		return err
+	}
+
+	dumpExec := dump.Exec(pgcommands.ExecOptions{StreamPrint: false})
+	if dumpExec.Error != nil {
+		s.l.Sugar().Fatalw("Failed to create database snapshot", "error", dumpExec.Error.Err, "output", dumpExec.Output)
```
seanmcgary: Fatalw will terminate the process which means your error wont actually return. This should be `.Errorw` instead to log the error

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	restore, err := s.setupRestore()
+	if err != nil {
+		return err
+	}
+
+	restoreExec := restore.Exec(s.cfg.InputFile, pgcommands.ExecOptions{StreamPrint: false})
+	if restoreExec.Error != nil {
+		s.l.Sugar().Fatalw("Failed to restore from snapshot",
```
seanmcgary: Same here - this should be `Errorw`

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
+	restore, err := pgcommands.NewRestore(&pgcommands.Postgres{
+		Host:     s.cfg.Host,
+		Port:     s.cfg.Port,
+		DB:       "", // left blank to not automatically assign DB as the role
+		Username: s.cfg.User,
+		Password: s.cfg.Password,
+	})
+	if err != nil {
+		s.l.Sugar().Fatalw("Failed to initialize restore", "error", err)
```
seanmcgary: Same here - needs to be `Errorw`

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if s.cfg.SchemaName != "" {
+		restore.SetSchemas([]string{s.cfg.SchemaName})
+	}
+
+	return restore, nil
+}
+
+// SnapshotConfigFromConfig converts a config.Config to a SnapshotConfig
+func SnapshotConfigFromConfig(cfg *config.Config) *SnapshotConfig {
```
seanmcgary: Going back to my previous comment, this couples the `config.go` module to this module, which we dont want. The snapshot config should be created inline wherever the SnapshotService is created.

seanmcgary (changes requested):

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,154 @@
...
+	return &SnapshotService{
+		cfg: cfg,
+		l:   l,
+	}
+}
+
+// CreateSnapshot creates a snapshot of the database based on the provided configuration.
+func (s *SnapshotService) CreateSnapshot() error {
+	if !s.validateCreateSnapshotConfig() {
+		return fmt.Errorf("invalid snapshot configuration")
```
seanmcgary: Going through and testing the CLI. Rather than having `validateCreateSnapshotConfig` log an error, it should return an error message on exactly what the issue is and let viper/cobra print it since we're using `RunE`. Otherwise, its going to be very easy for someone to miss what the error actually is since its json encoded currently:

image

seanmcgary:
Issue found with providing a output_file path of `~/snapshot-test`

```
go run main.go create-snapshot \
    --ethereum.rpc-url="..." \
    --chain="preprod" \
    --database.host="localhost" \
    --database.port="5432" \
    --database.user="seanmcgary" \
    --database.password="" \
    --database.db_name="sidecar_testnet_holesky" \
    --ethereum.use_native_batch_call=false \
    --output_file=~/snapshot-test
```

```
{"level":"error","ts":"2025-01-10T10:44:02.338-0600","caller":"snapshot/snapshot.go:50","msg":"Failed to create database snapshot","error":"exit status 1","output":"pg_dump: error: could not open output file \"~/snapshot-test\": No such file or directory\n","stacktrace":"github.com/Layr-Labs/sidecar/pkg/snapshot.(*SnapshotService).CreateSnapshot\n\t/Users/seanmcgary/Code/sidecar/pkg/snapshot/snapshot.go:50\ngithub.com/Layr-Labs/sidecar/cmd.init.func1\n\t/Users/seanmcgary/Code/sidecar/cmd/createSnapshot.go:37\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/seanmcgary/Code/gocode/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/seanmcgary/Code/gocode/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/seanmcgary/Code/gocode/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041\ngithub.com/Layr-Labs/sidecar/cmd.Execute\n\t/Users/seanmcgary/Code/sidecar/cmd/root.go:19\nmain.main\n\t/Users/seanmcgary/Code/sidecar/main.go:8\nruntime.main\n\t/opt/homebrew/opt/go/libexec/src/runtime/proc.go:272"}
Error: failed to create snapshot: exit status 1
```

seanmcgary:
Restoring a snapshot that was dumped using a schema (e.g. `public`) to a different schema, doesnt work but gives me a success message. New schema I specified ends up as completely empty


```bash
# implicitly uses the public schema
go run main.go create-snapshot \
    --ethereum.rpc-url="..." \
    --chain="preprod" \
    --database.host="localhost" \
    --database.port="5432" \
    --database.user="seanmcgary" \
    --database.password="" \
    --database.db_name="sidecar_testnet_holesky" \
    --ethereum.use_native_batch_call=false \
    --input_file=/Users/seanmcgary/snapshot-test

# specifies a schema that is not public
go run main.go restore-snapshot \
    --ethereum.rpc-url="..." \
    --chain="preprod" \
    --database.host="localhost" \
    --database.port="5432" \
    --database.user="seanmcgary" \
    --database.password="" \
    --database.db_name="sidecar_testnet_holesky" \
    --ethereum.use_native_batch_call=false \
    --input_file=/Users/seanmcgary/snapshot-test \
    --database.schema_name=test
```

BrendanCoughlan5:
> Restoring a snapshot that was dumped using a schema (e.g. `public`) to a different schema, doesnt work but gives me a success message. New schema I specified ends up as completely empty
> 
> ```shell
> # implicitly uses the public schema
> go run main.go create-snapshot \
>     --ethereum.rpc-url="..." \
>     --chain="preprod" \
>     --database.host="localhost" \
>     --database.port="5432" \
>     --database.user="seanmcgary" \
>     --database.password="" \
>     --database.db_name="sidecar_testnet_holesky" \
>     --ethereum.use_native_batch_call=false \
>     --input_file=/Users/seanmcgary/snapshot-test
> 
> # specifies a schema that is not public
> go run main.go restore-snapshot \
>     --ethereum.rpc-url="..." \
>     --chain="preprod" \
>     --database.host="localhost" \
>     --database.port="5432" \
>     --database.user="seanmcgary" \
>     --database.password="" \
>     --database.db_name="sidecar_testnet_holesky" \
>     --ethereum.use_native_batch_call=false \
>     --input_file=/Users/seanmcgary/snapshot-test \
>     --database.schema_name=test
> ```

More detail here:
https://www.notion.so/TDD-SideCar-Snapshots-14d13c11c3e0809c960cf1307d91d39e?d=17a13c11c3e0803c9ed6001cd067759d&pvs=4#17a13c11c3e0808e92a5e3800d1685bb

I thought about this case a fair bit,
I updated the CLI and docs accordingly and put in the docs how to convert the dump to a new schema.

cmd/createSnapshot.go
```diff
@@ -0,0 +1,60 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		initCreateSnapshotCmd(cmd)
+		cfg := config.NewConfig()
+
+		l, err := logger.NewLogger(&logger.LoggerConfig{Debug: cfg.Debug})
+		if err != nil {
+			return fmt.Errorf("failed to initialize logger: %w", err)
+		}
+
+		outputFile, err := utils.ExpandHomeDir(cfg.SnapshotConfig.OutputFile)
```
seanmcgary: 1. This should be happen as part of initializing the snapshot service or as part of the CreateSnapshot function
2. This helps to support referencing a home directory, but will still fail when someone enters a relative path

pkg/utils/helpers.go
```diff
@@ -24,3 +26,15 @@ func SnakeCase(s string) string {
 	notSnake := regexp.MustCompile(`[_-]`)
 	return notSnake.ReplaceAllString(s, "_")
 }
+
+// ExpandHomeDir expands the ~ in file paths to the user's home directory.
+func ExpandHomeDir(path string) (string, error) {
```
seanmcgary: There are functions within go that will do all of this for you automatically without having to write custom path logic. I encourage you to find them and use them (the `filepath` module is a great place to start)

scripts/convertSnapshotSchema.sh
```diff
@@ -0,0 +1,58 @@
...
+echo "Dropping temporary database if it exists: $TEMP_DB_NAME"
+PGPASSWORD=$DB_PASSWORD psql -U $DB_USER -c "DROP DATABASE IF EXISTS $TEMP_DB_NAME;" || { echo "Failed during DROP DATABASE IF EXISTS <temp_db_name>"; exit 1; }
+
+# Create a temporary database
+echo "Creating temporary database: $TEMP_DB_NAME"
+PGPASSWORD=$DB_PASSWORD psql -U $DB_USER -c "CREATE DATABASE $TEMP_DB_NAME;" || { echo "Failed to create database"; exit 1; }
+
+# Restore the snapshot dump into the temporary database
+echo "Restoring snapshot into temporary database"
+go run main.go restore-snapshot \
```
seanmcgary: This shouldnt ever reference `go run`, scripts meant to be used by users should always use the binary

scripts/convertSnapshotSchema.sh
```diff
@@ -0,0 +1,58 @@
...
+    --database.schema_name=$INPUT_SCHEMA_NAME \
+    --input_file=$INPUT_FILE || { echo "Failed to restore snapshot"; exit 1; }
+
+# Rename the schema in the temporary database
+echo "Renaming schema from $INPUT_SCHEMA_NAME to $OUTPUT_SCHEMA_NAME"
+PGPASSWORD=$DB_PASSWORD psql -U $DB_USER -d $TEMP_DB_NAME -c "ALTER SCHEMA $INPUT_SCHEMA_NAME RENAME TO $OUTPUT_SCHEMA_NAME;" || { echo "Failed to rename schema"; exit 1; }
+
+# Create a new snapshot with the updated schema
+echo "Creating new snapshot with updated schema"
+go run main.go create-snapshot \
```
seanmcgary: Same here, use the binary not go run

scripts/convertSnapshotSchema.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
seanmcgary, line 1: If we're adding this, why is this not part of the cli directly as a flag that the user can pass?

seanmcgary (changes requested):

BrendanCoughlan5:
btw i force pushed to rebase from master, no code change or conflict resolved in doing so

seanmcgary (approved):

PR #165, opened 2024-12-18
feat: add rewards-v2 support to staker-operator table generation

seanmcgary:


0xrajath (dismissed):
LGTM

pkg/rewards/stakerOperators/8_avsODStrategyPayouts.go
0xrajath, line 1: Just realised something. Should we dedupe this across strategies? I think in rewards calculation the staging step does a distinct that takes care of it.

0xrajath: Or do the same distinct filtering step in the stakerOperatorStaging step.

seanmcgary: We could make this a `sum(amount)` where its grouped by `avs, operator, strategy, reward_hash, snapshot`, but I think that should already be the case....

PR #178, opened 2025-01-09
feat: add leaf format forking logic for ODRewardSubmissions

seanmcgary:


pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
0xrajath, line 1: Just confirming that we don't need to do any of these changes for the default split model right?

internal/config/config.go
```diff
@@ -249,34 +257,62 @@ func (c *Config) GetGenesisBlockNumber() uint64 {
...
+		return ModelForkMap{
+			ModelFork_Austin: 3118055,
+		}, nil
+	case Chain_Holesky:
+		return ModelForkMap{
+			ModelFork_Austin: 3118055,
+		}, nil
+	case Chain_Mainnet:
+		return ModelForkMap{
+			ModelFork_Austin: 0, // doesnt apply to mainnet
```
0xrajath, line 311: Should this be a nil or -1 or something else? Since 0 is an actual block number

internal/config/config.go
```diff
@@ -249,34 +257,62 @@ func (c *Config) GetGenesisBlockNumber() uint64 {
...
 		return ForkMap{
-			Fork_Amazon:  "1970-01-01", // Amazon hard fork was never on preprod as we backfilled
-			Fork_Nile:    "2024-08-14", // Last calculation end timestamp was 8-13: https://holesky.etherscan.io/tx/0xb5a6855e88c79312b7c0e1c9f59ae9890b97f157ea27e69e4f0fadada4712b64#eventlog
-			Fork_Panama:  "2024-10-01",
-			Fork_Arno:    "2024-12-13",
-			Fork_Trinity: "2025-01-09",
+			RewardsFork_Amazon:  "1970-01-01", // Amazon hard fork was never on preprod as we backfilled
+			RewardsFork_Nile:    "2024-08-14", // Last calculation end timestamp was 8-13: https://holesky.etherscan.io/tx/0xb5a6855e88c79312b7c0e1c9f59ae9890b97f157ea27e69e4f0fadada4712b64#eventlog
+			RewardsFork_Panama:  "2024-10-01",
+			RewardsFork_Arno:    "2024-12-13",
```
0xrajath: We should revert this to whatever it was earlier. I think you had pushed some additional commits to the `testnet` branch right?

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
```diff
@@ -252,11 +252,43 @@ func (oas *OperatorAVSSplitModel) GenerateStateRoot(blockNumber uint64) ([]byte,
...
+	avs string,
+	activatedAt *time.Time,
+	oldOperatorAVSSplitBips uint64,
+	newOperatorAVSSplitBips uint64,
+) (string, error) {
+	modelForks, err := oas.globalConfig.GetModelForks()
+	if err != nil {
+		return "", err
+	}
+	if oas.globalConfig.ChainIsOneOf(config.Chain_Holesky, config.Chain_Preprod) && blockNumber < modelForks[config.ModelFork_Austin] {
```
0xrajath, line 270: Thinking out loud here. You don't need to check for specific chains, if you just define the block height correctly in the config right?

0xrajath: But I also guess it's fine to be explicit in every model.

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
```diff
@@ -252,11 +252,43 @@ func (oas *OperatorAVSSplitModel) GenerateStateRoot(blockNumber uint64) ([]byte,
...
 		slotID := base.NewSlotID(split.TransactionHash, split.LogIndex)
-		value := fmt.Sprintf("%s_%s_%016x_%016x_%016x", split.Operator, split.Avs, split.ActivatedAt.Unix(), split.OldOperatorAVSSplitBips, split.NewOperatorAVSSplitBips)
+		value, err := oas.formatMerkleLeafValue(split.BlockNumber, split.Operator, split.Avs, split.ActivatedAt, split.OldOperatorAVSSplitBips, split.NewOperatorAVSSplitBips)
+		if err != nil {
+			oas.logger.Sugar().Errorw("Failed to format merkle leaf value",
+				zap.Error(err),
+				zap.Uint64("blockNumber", split.BlockNumber),
+				zap.String("operator", split.Operator),
+				zap.String("avs", split.Avs),
+				zap.Time("activatedAt", *split.ActivatedAt),
```
0xrajath, line 289: Still catching up on golang semantics. Anything weird with having it as `*split.ActivatedAt` (i.e by Reference) while the others are not?

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
```diff
@@ -252,11 +252,43 @@ func (oas *OperatorAVSSplitModel) GenerateStateRoot(blockNumber uint64) ([]byte,
...
+				zap.Uint64("oldOperatorAVSSplitBips", split.OldOperatorAVSSplitBips),
+				zap.Uint64("newOperatorAVSSplitBips", split.NewOperatorAVSSplitBips),
+			)
+			continue
```
0xrajath: Why do we continue the loop here? It should just fail if it's not able to format correctly right?

pkg/eigenState/operatorDirectedRewardSubmissions/operatorDirectedRewardSubmissions.go
```diff
@@ -80,6 +76,14 @@ func (odrs *OperatorDirectedRewardSubmissionsModel) GetModelName() string {
 	return "OperatorDirectedRewardSubmissionsModel"
 }
 
+func (odrs *OperatorDirectedRewardSubmissionsModel) NewSlotID(blockNumber uint64, transactionHash string, logIndex uint64, rewardHash string, strategyIndex uint64, operatorIndex uint64) types.SlotID {
+	if odrs.globalConfig.ChainIsOneOf(config.Chain_Holesky, config.Chain_Preprod) && blockNumber < 3118055 {
```
0xrajath: We should be pulling the block number from the config rather than hardcoding IMO.

pkg/eigenState/operatorDirectedRewardSubmissions/operatorDirectedRewardSubmissions.go
```diff
@@ -313,24 +317,58 @@ func (odrs *OperatorDirectedRewardSubmissionsModel) GenerateStateRoot(blockNumbe
...
+	multiplier string,
+	operator string,
+	amount string,
+) (string, error) {
+	modelForks, err := odrs.globalConfig.GetModelForks()
+	if err != nil {
+		return "", err
+	}
+
+	if odrs.globalConfig.ChainIsOneOf(config.Chain_Holesky, config.Chain_Preprod) && blockNumber < modelForks[config.ModelFork_Austin] {
```
0xrajath, line 356: Since this condition is reused a couple of times in this Model, does it make sense to refactor it as a function and then just call that function to check?

pkg/eigenState/operatorPISplits/operatorPISplits.go
```diff
@@ -250,11 +250,41 @@ func (ops *OperatorPISplitModel) GenerateStateRoot(blockNumber uint64) ([]byte,
...
+		if err != nil {
+			ops.logger.Sugar().Errorw("Failed to format merkle leaf value",
+				zap.Error(err),
+				zap.Uint64("blockNumber", split.BlockNumber),
+				zap.String("operator", split.Operator),
+				zap.Time("activatedAt", *split.ActivatedAt),
+				zap.Uint64("oldOperatorPISplitBips", split.OldOperatorPISplitBips),
+				zap.Uint64("newOperatorPISplitBips", split.NewOperatorPISplitBips),
+			)
+			continue
```
0xrajath: It should halt rather than continue if there is an error right?

internal/config/config.go
seanmcgary: Oh yeah, lemme do that...

internal/config/config.go
seanmcgary: I'll make it `-1` to make it a little more clear since ints cant be nil...

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
seanmcgary: Shouldnt have to change anything here since the default split value is only used when looking at the trinity fork. Totally fine to compile the data if it exists before then since it wont get used till the fork date.

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
seanmcgary: No because mainnet and holesky have overlapping block counts. Would rather be SUPER explicit to say what chain the fork applies to

pkg/eigenState/operatorAVSSplits/operatorAVSSplits.go
seanmcgary: good catch on this, it shouldnt have the pointer dereference. I'll make sure to file a complaint with Copilot 😂

seanmcgary: Actually this is correct - `zap.Time` takes a value not a pointer so dereferencing is correct

pkg/eigenState/operatorDirectedRewardSubmissions/operatorDirectedRewardSubmissions.go
seanmcgary: Theres not really a clean way to do it that saves you anything. you'd end up with basically a function that checks the condition and also takes an anonymous function or something, which isnt really any better.

internal/config/config.go
```diff
@@ -249,34 +257,62 @@ func (c *Config) GetGenesisBlockNumber() uint64 {
...
+
+func (c *Config) GetModelForks() (ModelForkMap, error) {
+	switch c.Chain {
+	case Chain_Preprod:
+		return ModelForkMap{
+			ModelFork_Austin: 3118055,
+		}, nil
+	case Chain_Holesky:
+		return ModelForkMap{
+			ModelFork_Austin: 3118055,
```
0xrajath: Should we update these block heights to be less than 22 hours out?

seanmcgary: yup, was just waiting for everything else to be approved

0xrajath (dismissed):
LGTM

PR #182, opened 2025-01-13
feat: add rewards-v2 mainnet ABIs

seanmcgary:


0xrajath (approved):

pkg/contractStore/coreContracts/mainnet.json
```diff
@@ -74,6 +74,10 @@
...
 			"bytecode_hash": "8e37325a8c09555be3e2ec19453cc67ee202f6847cc5348b263896126d7268a8"
+		}, {
+			"contract_address": "0x29a954e9e7f12936db89b183ecdf879fbbb99f14",
```
0xrajath, line 78: Shouldn't this be the proxy address (i.e the RewardsCoordinator) and not the implementation?

0xrajath: Oh or this the rolling up of all implementations you mentioned?

seanmcgary: Nope, that association is below. This is the address of the implementation and its ABI

PR #189, opened 2025-01-15
fix: add postgresql client to docker

BrendanCoughlan5:
adds postgresql-client to the docker build for pg_dump and pg_restore which https://github.com/habx/pg-commands uses for snapshot create and snapshot restore


seanmcgary (approved):

BrendanCoughlan5:
ty :)

PR #201, opened 2025-01-27
feat: hash checksums + download urls for snapshot restore and create

BrendanCoughlan5:
hash checksums + download urls for snapshot restore and create, also cleaned up/ logs/ readme/ docs


README.md
```diff
@@ -125,8 +125,15 @@ Run the sidecar
...
 
+  Note: This command restores --database.schema_name only if it's present in input snapshot file.
+  Follow the snapshot docs if you need to convert the snapshot to a different schema name than was used during snapshot creation. sidecar.eigenlayer.xyz snapshots are only served in the public schema.
+
 Flags:
   -h, --help   help for run
+      --snapshot.input_file string   Path to the snapshot file (can override the metadata_url latest snapshot with an input_file or input_url)
+      --snapshot.input_url string   URL to the snapshot file (can override the metadata_url latest snapshot with an input_file or input_url)
+      --snapshot.hash_file string  Path to save the snapshot file to (can override the metadata_url to get a snapshot hash with a hash_file)
+      --snapshot.hash_url string  Path to save the snapshot file to (can override the metadata_url to get a snapshot hash with a hash_url)
```
seanmcgary: Rather than polluting the cli with so many variable options, I would have two flags:

```
--snapshot.input (string)
--snapshot.verify-input  (bool)
```

`snapshot.input` should be smart: if the value is a correctly formed URL, it should treat it as such and download the file. If its not a url, it should be treated as a file on the local filesystem.

If `verify-input` is true, the sidecar should look for a file with the same name as the input but with a `.sha256sum` extension (assuming we're using `sha256sum` to generate the checksum). If its a URL, download it.

docs/snapshots_docs.md
```diff
@@ -64,8 +64,11 @@ Usage:
   sidecar restore-snapshot [flags]
 
 Flags:
-  -h, --help                help for restore-snapshot
-      --input_file string   Path to the snapshot file (required)
+  -h, --help                     help for restore-snapshot
+      --input_file string        Path to the snapshot file (can override the metadata_url latest snapshot with an input_file or input_url)
```
seanmcgary: This is a bit confusing when reading the `--help` menu and trying to figure out what these mean.

First I would flag this as optional since as-written it seems to be required.

Second, I would get more explicit with it. Something like `--snapshot.override-input=`

docs/snapshots_docs.md
```diff
@@ -64,8 +64,11 @@ Usage:
...
+      --input_hash_file string   Path to the hash file
+      --input_hash_url string    URL to the hash file (overrides the hash_file)
+      --input_url string         URL to the snapshot file (can override the metadata_url latest snapshot with an input_file or input_url)
```
seanmcgary: See my comment at the top. Hash file names should be inferred from the name of the file with a known extension (e.g. `.sha256sum`)

pkg/snapshot/snapshot.go
```diff
@@ -12,25 +16,47 @@ import (
...
+	l         *zap.Logger
+	tempFiles []string
 }
 
 // NewSnapshotService initializes a new SnapshotService with the given configuration and logger.
 func NewSnapshotService(cfg *SnapshotConfig, l *zap.Logger) (*SnapshotService, error) {
 	var err error
+	tempFiles := []string{}
+
+	if cfg.InputFile == "" && cfg.InputURL != "" {
```
seanmcgary: Related to my very top comment: we should have a single input flag that can take a url OR a file path and the SnapshotService should sort out what to do (e.g. download or not)

pkg/snapshot/snapshot.go
```diff
@@ -40,12 +66,21 @@ func NewSnapshotService(cfg *SnapshotConfig, l *zap.Logger) (*SnapshotService, e
...
 		return nil, fmt.Errorf("failed to resolve output file path: %w", err)
 	}
+	cfg.InputHashFile, err = resolveFilePath(cfg.InputHashFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve input hash file path: %w", err)
+	}
+	cfg.OutputHashFile, err = resolveFilePath(cfg.OutputHashFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve output hash file path: %w", err)
+	}
```
seanmcgary: Can we clean up the body of this function a bit? It looks like we're doing stuff for both create and restore functions here. e.g., running `NewSnapshotService` for a restore shouldnt care what the output hash file is

pkg/snapshot/snapshot.go
```diff
@@ -40,12 +66,21 @@ func NewSnapshotService(cfg *SnapshotConfig, l *zap.Logger) (*SnapshotService, e
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve output file path: %w", err)
 	}
+	cfg.InputHashFile, err = resolveFilePath(cfg.InputHashFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve input hash file path: %w", err)
+	}
+	cfg.OutputHashFile, err = resolveFilePath(cfg.OutputHashFile)
```
seanmcgary: Related to my above comments, theres no need to take a specific param for the output hash file file. This should be the output file name with `.sha256sum` appended as an extension.

pkg/snapshot/snapshot.go
```diff
@@ -40,12 +66,21 @@ func NewSnapshotService(cfg *SnapshotConfig, l *zap.Logger) (*SnapshotService, e
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve output hash file path: %w", err)
+	}
 
-	l.Sugar().Infow("Resolved file paths", "inputFile", cfg.InputFile, "outputFile", cfg.OutputFile)
+	l.Sugar().Infow("Resolved file paths", "inputFile", cfg.InputFile, "outputFile", cfg.OutputFile, "inputHashFile", cfg.InputHashFile, "outputHashFile", cfg.OutputHashFile)
```
seanmcgary: This should likely be a `.Debug()` log level.

Can we also fix the formatting to be multi-line and use `zap.<type>("key", <value>)` like we do elsewhere in the codebase for consistency?

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +231,89 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+		return nil, fmt.Errorf("failed to open file: %w", err)
+	}
+	defer file.Close()
+
+	hasher := sha256.New()
+	if _, err := io.Copy(hasher, file); err != nil {
+		return nil, fmt.Errorf("failed to hash file: %w", err)
+	}
+
+	return hasher.Sum(nil), nil
```
seanmcgary, line 322: Let's change the format of this to match the output from the `sha256sum` cli utility. The value returned that gets stored in the resulting file should be:

```
<hash> <filename>\n
```

e.g.
```
6806a717f2610b3f70f2228fc67f7b6ffe96ce1576e201ac7c328a55779f1bdc  README.md
```

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +231,89 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
 
 	return restore, nil
 }
+
+func (s *SnapshotService) saveOutputFileHash() error {
```
seanmcgary: For the sake of style and testability, can we make this more of a "pure" function that takes as arguments:

* The name/path of the file to hash
* The place where to put it

Doing so will make this function far more testable and not reliant on internal state.

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +231,89 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+	if err := os.WriteFile(s.cfg.OutputHashFile, []byte(hashString), 0644); err != nil {
+		return fmt.Errorf("failed to write hash to file: %w", err)
+	}
+
+	s.l.Sugar().Infow("Output file hash saved successfully", "outputFile", s.cfg.OutputFile, "outputHashFile", s.cfg.OutputHashFile)
+	return nil
+}
+
+// Cleanup removes any temporary files created during the snapshot process.
+func (s *SnapshotService) Cleanup() {
```
seanmcgary: Same idea here: `Cleanup()` should take a slice of files to cleanup as an argument rather than relying on internal state of the struct. Makes it more testable. You could also go one step further and take the logger instance as an argument and decouple this function from the struct all together.

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +231,89 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to download file: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("failed to download file: received status code %d", resp.StatusCode)
+	}
+
+	tmpFile, err := os.CreateTemp("", fmt.Sprintf("%s-*.tmp", prefix))
```
seanmcgary: Why do we need to suffix the file with `.tmp`?

pkg/snapshot/snapshot_test.go
```diff
@@ -79,24 +84,32 @@ func TestSetupSnapshotDump(t *testing.T) {
...
 	snapshotFile := filepath.Join(tempDir, "TestValidateRestoreConfig.sql")
-	_, err := os.Create(snapshotFile)
-	assert.NoError(t, err, "Creating snapshot file should not fail")
+	snapshotHashFile := filepath.Join(tempDir, "TestValidateRestoreConfig.hash")
+	content := []byte("test content")
+	err := os.WriteFile(snapshotFile, content, 0644)
+	assert.NoError(t, err, "Writing to snapshot file should not fail")
+
+	hash := sha256.Sum256(content)
+	err = os.WriteFile(snapshotHashFile, []byte(hex.EncodeToString(hash[:])), 0644)
```
seanmcgary: Recommend using the `saveOutputFileHash()` you created rather than hashing the file manually.

seanmcgary (changes requested):

docs/snapshots_docs.md
BrendanCoughlan5: updated for --snapshot.input for now, and make it required, will change depending on the metadata with desired design

docs/snapshots_docs.md
BrendanCoughlan5: now inferring from name

README.md
BrendanCoughlan5: changed for these flags

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed to be just the input now instead of url and path

pkg/snapshot/snapshot.go
BrendanCoughlan5: put the logic in restore and create now :)

pkg/snapshot/snapshot.go
BrendanCoughlan5: done

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to use debug

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to pure

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to pure

pkg/snapshot/snapshot_test.go
BrendanCoughlan5: switched to use saveOutputFileHash()

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed, i think it's a little more readable but NIT

docs/src/content/docs/running/snapshots.md
```diff
@@ -5,23 +5,159 @@ description: How to use a snapshot to start or restore your Sidecar
...
+The input can be a local file path or a URL, url of type http, https, or ftp is supported.
+Follow the snapshot docs if you need to convert the snapshot to a different schema name than was used during snapshot creation.
+
+Usage:
+  sidecar restore-snapshot [flags]
+
+Flags:
+  -h, --help                    help for restore-snapshot
+      --snapshot.input string   Path to the snapshot file either a URL or a local file (required)
+      --snapshot.verify_input   Boolean to verify the input file against its .sha256sum file, if input is a url then it downloads the file, (default is true) (default true)
```
seanmcgary: Im not sure how this happened, but it looks like we have some consistency drift in the formatting of arguments.

For example, `--rpc.http-port` is the correct formatting with dashes. Can we update `--snapshot.verify_input` to be `--snapshot.verify-input`? I'll check on the others that are not related to snapshots to see what happened there...

internal/config/config.go
```diff
@@ -126,8 +127,9 @@ var (
 	DatabaseDbName     = "database.db_name"
 	DatabaseSchemaName = "database.schema_name"
 
-	SnapshotOutputFile = "output_file"
-	SnapshotInputFile  = "input_file"
+	SnapshotOutputFile  = "snapshot.output_file"
+	SnapshotInput       = "snapshot.input"
+	SnapshotVerifyInput = "snapshot.verify_input"
```
seanmcgary: Related to my comment above, but the underscores should be dashes.

pkg/snapshot/snapshot.go
```diff
@@ -1,52 +1,63 @@
...
 	}
-
-	l.Sugar().Infow("Resolved file paths", "inputFile", cfg.InputFile, "outputFile", cfg.OutputFile)
-
-	return &SnapshotService{
-		cfg: cfg,
-		l:   l,
-	}, nil
+	// Allow HTTP, HTTPS, and FTP URL schemes
+	return u.Scheme == "http" || u.Scheme == "https" || u.Scheme == "ftp"
```
seanmcgary: I would only allow http(s) here. ftp is not in scope

pkg/snapshot/snapshot.go
```diff
@@ -70,6 +81,15 @@ func resolveFilePath(path string) (string, error) {
...
 // CreateSnapshot creates a snapshot of the database based on the provided configuration.
 func (s *SnapshotService) CreateSnapshot() error {
+	var err error
+
+	s.cfg.OutputFile, err = resolveFilePath(s.cfg.OutputFile)
+	if err != nil {
+		return fmt.Errorf("failed to resolve output file path: %w", err)
+	}
+
+	s.l.Debug("Resolved file paths", zap.String("Output", s.cfg.OutputFile))
```
seanmcgary: Should be `s.l.Sugar().Debugw()`

pkg/snapshot/snapshot.go
```diff
@@ -70,6 +81,15 @@ func resolveFilePath(path string) (string, error) {
 
 // CreateSnapshot creates a snapshot of the database based on the provided configuration.
 func (s *SnapshotService) CreateSnapshot() error {
+	var err error
+
+	s.cfg.OutputFile, err = resolveFilePath(s.cfg.OutputFile)
+	if err != nil {
+		return fmt.Errorf("failed to resolve output file path: %w", err)
```
seanmcgary: To make this more meaningful, I would do something like:

```go
return errors.Wrap(err, fmt.Sprintf("failed to resolve output file path '%s', s.cfg.OutputFile))
```

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +148,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
 
+// resolveSnapshotServiceInput resolves the SnapshotService struct into a suitable format for restoring a snapshot
+func (s *SnapshotService) resolveRestoreInput() error {
+	var err error
+
+	if isNetworkURL(s.cfg.Input) {
+		inputUrl := s.cfg.Input
+
+		uniqueID := fmt.Sprintf("%d", time.Now().UnixNano())
+		inputFileName := uniqueID + "_downloaded_snapshot.dump"
```
seanmcgary: why are we adding additional stuff to the file name rather than just using the name of the file as it exists in the bucket?

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +148,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to download snapshot: %w", err)
+		}
+		s.tempFiles = append(s.tempFiles, s.cfg.Input)
+
+		if s.cfg.VerifyInput {
+			hashFileName := inputFileName + ".sha256sum"
+			hashFile, err := downloadFile(inputUrl+".sha256sum", hashFileName)
+			if err != nil {
+				return fmt.Errorf("failed to download snapshot hash: %w", err)
```
seanmcgary: Do we want to differentiate between an problem downloading the file and an error like a 404 where the file simply doesnt exist? If it doesnt exist, what do you think is the best thing to do?

pkg/snapshot/snapshot.go
```diff
@@ -148,13 +222,23 @@ func (s *SnapshotService) setupSnapshotDump() (*pgcommands.Dump, error) {
...
 
-	info, err := os.Stat(s.cfg.InputFile)
+	info, err := os.Stat(s.cfg.Input)
 	if err != nil || info.IsDir() {
-		return fmt.Errorf("snapshot file does not exist: %s", s.cfg.InputFile)
+		return fmt.Errorf("snapshot file does not exist: %s", s.cfg.Input)
+	}
+
+	if s.cfg.VerifyInput {
+		if err := validateInputFileHash(s.cfg.Input, s.cfg.Input+".sha256sum"); err != nil {
```
seanmcgary: We do a lot of concatenation of file name to `sha256sum` all over the place. Can we DRY this up a bit?

pkg/snapshot/snapshot_test.go
```diff
@@ -134,6 +152,105 @@ func TestSetupRestore(t *testing.T) {
...
+			t.Logf("Failed to remove downloaded file: %v", err)
+		}
+	}()
+
+	// Verify the file was downloaded correctly
+	content, err := os.ReadFile(filePath)
+	assert.NoError(t, err, "Reading downloaded file should not fail")
+	assert.Contains(t, string(content), "This is a test file.", "Downloaded file content should match expected content")
+}
+
```
seanmcgary, line 242: We also need to add a test to make sure the generated hash will work with the `sha256sum` cli in bash.

seanmcgary (changes requested):

docs/src/content/docs/running/snapshots.md
BrendanCoughlan5: i noticed that as well, i can change it to verify-input, there's no dependance on verify-input at the moment so that's fine

BrendanCoughlan5: I'll just make the snapshot ones verify_input and output_file to use the - instead of _

internal/config/config.go
BrendanCoughlan5: I'll just make the snapshot ones verify_input and output_file to use the - instead of _

pkg/snapshot/snapshot.go
BrendanCoughlan5: avoids name collision, i.e. an already fully downloaded file being accidentally removed in a cleanup / you don't have to worry about weird custom URL's that people want to make and use. trying to clean url's is a bit messy

pkg/snapshot/snapshot.go
BrendanCoughlan5: When using the repo_metadata/ manifest, if there's a 404 and you can't find the file, then you should go through older versions, i.e. curl -I through newest -> oldest, then download it, i think this would be handled no in the downloadFile but in a function interpreting the manifest to find the file.

pkg/snapshot/snapshot.go
BrendanCoughlan5: i switched it to be opinionated that you need a base path

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed :)

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to use wrap

pkg/snapshot/snapshot.go
BrendanCoughlan5: adjusted

pkg/snapshot/snapshot.go
BrendanCoughlan5: also did the errors.Wrap and the downloadFile gives a 404 to be more explicit

pkg/snapshot/snapshot.go
BrendanCoughlan5: done

pkg/snapshot/snapshot_test.go
BrendanCoughlan5: done :)

pkg/snapshot/snapshot.go
BrendanCoughlan5: potentially you could give an error that the hash file isn't there, no verification happened and then proceed with restoring the snapshot, but i think that might be dangerous/ security wouldn't like that? Maybe for testnet that's fine to allow, i'm a bit hesitant to allow restore snapshot without

docs/src/content/docs/running/snapshots.md
```diff
@@ -5,23 +5,159 @@ description: How to use a snapshot to start or restore your Sidecar
...
+  --database.user=sidecar \
+  --database.password=sidecar \
+  --database.port=5432 \
+  --database.db_name=sidecar \
+  --database.schema_name=public \
+  --snapshot.output-file=example.dump
+```
+
+
+## Advanced: Converting the Schema of a Dump
```
seanmcgary: We should do one of a few things here:

1. Omit this entirely since it's an edgecase that the vast majority of people wont actually need
2. If we are going to include it, the script itself needs to be cleaned up and made production ready. Right now it makes a lot of assumptions such as host, port, etc. The person running the script should be able to configure all of those values at runtime.
3. Option 2 plus put this section in a new section (so new file under the `/running` directory) specific to advanced usage.

docs/src/content/docs/running/snapshots.md
```diff
@@ -5,23 +5,159 @@ description: How to use a snapshot to start or restore your Sidecar
...
+# Can use the script
+./scripts/convertSnapshotSchema.sh 
+
+```bash
+./scripts/convertSnapshotSchema.sh <inputSchema> <outputSchema> input_file.dump output_file.dump <db_username> <db_password>
+```
+
+
+```bash
+# Open your terminal and create a temporary database to work with:
```
seanmcgary: Wait, do I need to run the following if I ran the `convertSnapshotSchema.sh` script already? or is this just a different way to achieve the same thing?

pkg/snapshot/snapshot.go
```diff
@@ -1,52 +1,64 @@
...
+		return false
 	}
-
-	l.Sugar().Infow("Resolved file paths", "inputFile", cfg.InputFile, "outputFile", cfg.OutputFile)
-
-	return &SnapshotService{
-		cfg: cfg,
-		l:   l,
-	}, nil
+	// Allow HTTP, HTTPS, and FTP URL schemes
```
seanmcgary: except it doesn't allow FTP. IMO, Id remove this comment entirely since its fairly obvious by the if-condition what is allowed.

pkg/snapshot/snapshot.go
```diff
@@ -70,6 +82,15 @@ func resolveFilePath(path string) (string, error) {
 
 // CreateSnapshot creates a snapshot of the database based on the provided configuration.
 func (s *SnapshotService) CreateSnapshot() error {
+	var err error
```
seanmcgary: stylistically, no need to declare `err` up here. it can be declared and assigned in the following statement

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +149,47 @@ func (s *SnapshotService) RestoreSnapshot() error {
 	return nil
 }
 
+// resolveSnapshotServiceInput resolves the SnapshotService struct into a suitable format for restoring a snapshot
+func (s *SnapshotService) resolveRestoreInput() error {
+	var err error
```
seanmcgary: same as above, why are we declaring `err` here at the top?

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +149,47 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+
+	if isNetworkURL(s.cfg.Input) {
+		inputUrl := s.cfg.Input
+
+		// Use getFileNameFromURL to extract the file name
```
seanmcgary: function name is already self-explanatory, would recommend removing the comment

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +149,47 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+		fileName, err := getFileNameFromURL(inputUrl)
+		if err != nil {
+			return fmt.Errorf("failed to extract file name from URL: %w", err)
+		}
+
+		// Use a temporary directory for the download
+		inputFilePath := filepath.Join("tmp", fileName)
```
seanmcgary: This should use `os.TempDir()` as the temp directory, otherwise this is going to create a `tmp` directory in whatever directory the user happened to run the command in.

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +149,47 @@ func (s *SnapshotService) RestoreSnapshot() error {
 	return nil
 }
 
+// resolveSnapshotServiceInput resolves the SnapshotService struct into a suitable format for restoring a snapshot
+func (s *SnapshotService) resolveRestoreInput() error {
```
seanmcgary: Can rename this to be more accurate?

> resolves the SnapshotService struct into a suitable format for restoring a snapshot

Yes it does this, but also downloads the file, which is not apparent in the function name or documentation for the function

pkg/snapshot/snapshot.go
```diff
@@ -148,13 +225,23 @@ func (s *SnapshotService) setupSnapshotDump() (*pgcommands.Dump, error) {
...
 
 func (s *SnapshotService) validateRestoreConfig() error {
-	if s.cfg.InputFile == "" {
-		return fmt.Errorf("restore snapshot file path i.e. `input-file` must be specified")
+	if s.cfg.Input == "" {
+		return fmt.Errorf("restore snapshot file path i.e. `input` must be specified")
 	}
 
-	info, err := os.Stat(s.cfg.InputFile)
+	info, err := os.Stat(s.cfg.Input)
```
seanmcgary: What if the input is a URL to the dump to be restored?

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +269,125 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+		return "", errors.Wrap(err, "failed to initiate download")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("downloading error, received status code %d", resp.StatusCode)
+	}
+
+	// Ensure the directory for the file path exists
+	dir := filepath.Dir(filePath)
```
seanmcgary: Shouldnt we actually make sure the directory exists first before running `MkdirAll`?

Also you have `filepath` (the module) and `filePath` which is actually really confusing. Can we rename `filePath` to `downloadDestFilePath` or something?

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +269,125 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+		return "", errors.Wrap(err, "failed to create directories to save the download file to")
+	}
+
+	// For CLI UX Get the content length for the progress bar
+	contentLength := resp.ContentLength
```
seanmcgary: no need for the intermediate assignment; you can just use `resp.ContentLength` in the progressbar call

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +269,125 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+		progressbar.OptionSetDescription(fmt.Sprintf("downloading %s", filePath)),
+		progressbar.OptionSetTheme(progressbar.Theme{Saucer: "=", SaucerPadding: " ", BarStart: "[", BarEnd: "]"}),
+	)
+
+	tmpFile, err := os.Create(filePath)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create local file")
+	}
+
+	// Use io.TeeReader to update the progress bar while copying
```
seanmcgary: except we're not using io.TeeReader. recommend removing outdated comment

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +269,125 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+	_, err = io.Copy(io.MultiWriter(tmpFile, bar), resp.Body)
+	if err != nil {
+		tmpFile.Close()
+		return "", errors.Wrap(err, "failed to write to local file")
+	}
+
+	tmpFile.Close()
```
seanmcgary: this should be moved to the line under where tmpFile is created and needs a `defer` in front of it to make sure it gets closed properly even if we return early.

pkg/snapshot/snapshot.go
```diff
@@ -182,3 +269,125 @@ func (s *SnapshotService) setupRestore() (*pgcommands.Restore, error) {
...
+	fileContent := fmt.Sprintf("%s %s\n", hex.EncodeToString(hash), filepath.Base(filePath))
+	if err := os.WriteFile(hashFilePath, []byte(fileContent), 0644); err != nil {
+		return fmt.Errorf("failed to write hash to file: %w", err)
+	}
+
+	return nil
+}
+
+// cleanup removes the specified files and logs the results using the provided logger.
+func cleanup(files []string, l *zap.Logger) {
```
seanmcgary: Can we make this more specific? Something like `cleanupTempFiles`?

pkg/snapshot/snapshot.go
```diff
@@ -81,16 +102,31 @@ func (s *SnapshotService) CreateSnapshot() error {
...
+	if err := saveOutputFileHash(s.cfg.OutputFile, outputHashFile); err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to save output file hash '%s'", outputHashFile))
+	}
+
 	return nil
 }
 
 // RestoreSnapshot restores a snapshot of the database based on the provided configuration.
 func (s *SnapshotService) RestoreSnapshot() error {
+	defer cleanup(s.tempFiles, s.l)
```
seanmcgary: We delete the temp files, but we dont ever cleanup the list of temp files on the struct (we should)

seanmcgary (approved):

seanmcgary (changes requested):
Selected the wrong option on accident...see above feedback for changes needed.

internal/config/config.go
```diff
@@ -126,8 +127,9 @@ var (
 	DatabaseDbName     = "database.db_name"
 	DatabaseSchemaName = "database.schema_name"
 
-	SnapshotOutputFile = "output_file"
-	SnapshotInputFile  = "input_file"
+	SnapshotOutputFile  = "snapshot.output-file"
+	SnapshotInput       = "snapshot.input"
+	SnapshotVerifyInput = "snapshot.verify-input"
```
seanmcgary, line 132: Since these are scoped to the create/restore commands already, why did we re-add the `snapshot` prefix?

seanmcgary:
Ran the `restore-snapshot` function and encountered a strange error. It seems to have downloaded the file completely, but then say's it received a 404 after the fact (which isnt possible):

```
downloading tmp/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump 100% [========================================] Error: failed to restore snapshot: Error downloading snapshot from 'https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump': downloading error, received status code 404
```

It also didnt clean up the file that was created, which it probably should.

arguments:

```bash
go run main.go restore-snapshot
    --ethereum.rpc-url="https://winter-white-crater.ethereum-holesky.quiknode.pro/1b1d75c4ada73b7ad98e1488880649d4ea637733/" \
    --chain="preprod" \
    --database.host="localhost" \
    --database.port="5432" \
    --database.user="seanmcgary" \
    --database.password="" \
    --database.db_name="sidecar_test_restore" \
        --ethereum.use_native_batch_call=false \
        --snapshot.input=https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump \
        --database.schema_name=public
```

As a sidenote, when we encounter an error during execution, can we change it so that the usage message isnt printed? It adds a lot of noise and should really only be printed if the inputs are incorrect.

docs/src/content/docs/running/snapshots.md
BrendanCoughlan5: i'm intending of removing it once we migrate to protocol-sidecar and use the public schema for testnet-holesky, in the meantime can omit it if you prefer

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed the comment

pkg/snapshot/snapshot.go
BrendanCoughlan5: updated to restoreAndDownloadRestoreInput

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to do 

	resolvedFilePath, err := resolveFilePath(s.cfg.OutputFile)
	if err != nil {
		return errors.Wrap(err, fmt.Sprintf("failed to resolve output file path '%s'", s.cfg.OutputFile))
	}
	s.cfg.OutputFile = resolvedFilePath

pkg/snapshot/snapshot.go
BrendanCoughlan5: now cleans up the struct as well

pkg/snapshot/snapshot.go
BrendanCoughlan5: switched to this

	resolvedFilePath, err := resolveFilePath(s.cfg.Input)
	if err != nil {
		return errors.Wrap(err, fmt.Sprintf("failed to resolve input file path '%s'", s.cfg.Input))
	}
	s.cfg.Input = resolvedFilePath

pkg/snapshot/snapshot.go
BrendanCoughlan5: now uses os.TempDir()

pkg/snapshot/snapshot.go
BrendanCoughlan5: s.cfg.Input, err = downloadFile(inputUrl, inputFilePath) in the prepareAndDownloadRestoreInput() function
s.cfg.Input is a filepath at this point

BrendanCoughlan5: put a comment above to make it clearer

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed to be cleanupTempFiles

pkg/snapshot/snapshot.go
BrendanCoughlan5: updated to downloadDestFilePath

BrendanCoughlan5: // MkdirAll creates a directory named path,
// along with any necessary parents, and returns nil,
// or else returns an error.
// The permission bits perm (before umask) are used for all
// directories that MkdirAll creates.
// If path is already a directory, MkdirAll does nothing
// and returns nil.

So it's fine in this language, but made it if you prefer/ find it more readable

	if _, err := os.Stat(dir); os.IsNotExist(err) {
		if err := os.MkdirAll(dir, os.ModePerm); err != nil {
			return "", errors.Wrap(err, "failed to create directories to save the download file to")
		}
	}

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed

pkg/snapshot/snapshot.go
BrendanCoughlan5: moved

docs/src/content/docs/running/snapshots.md
BrendanCoughlan5: just gave the restore and create method instead

docs/src/content/docs/running/snapshots.md
BrendanCoughlan5: actually, just remove the script entirely from the docs, just get them to do the restore-snapshot and create snapshot method if they need to do it

internal/config/config.go
BrendanCoughlan5: consistency for the upcoming one click/ run, so you only need to do the environment variable once to use between them.

 https://www.notion.so/TDD-SideCar-Snapshots-14d13c11c3e0809c960cf1307d91d39e?d=18f13c11c3e08097926e001cc4982f37&pvs=4#18a13c11c3e080b3a4f6c34be7dc923b

BrendanCoughlan5:
> Ran the `restore-snapshot` function and encountered a strange error. It seems to have downloaded the file completely, but then say's it received a 404 after the fact (which isnt possible):
> 
> ```
> downloading tmp/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump 100% [========================================] Error: failed to restore snapshot: Error downloading snapshot from 'https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump': downloading error, received status code 404
> ```
> 
> It also didnt clean up the file that was created, which it probably should.
> 
> arguments:
> 
> ```shell
> go run main.go restore-snapshot
>     --ethereum.rpc-url="https://winter-white-crater.ethereum-holesky.quiknode.pro/1b1d75c4ada73b7ad98e1488880649d4ea637733/" \
>     --chain="preprod" \
>     --database.host="localhost" \
>     --database.port="5432" \
>     --database.user="seanmcgary" \
>     --database.password="" \
>     --database.db_name="sidecar_test_restore" \
>         --ethereum.use_native_batch_call=false \
>         --snapshot.input=https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump \
>         --database.schema_name=public
> ```
> 
> As a sidenote, when we encounter an error during execution, can we change it so that the usage message isnt printed? It adds a lot of noise and should really only be printed if the inputs are incorrect.

Fixed to be clear it's the filesum the hash that doesn't exist. 


Try with the --snapshot.verify-input=false flag





It should now give something like  : {"level":"fatal","ts":"2025-02-03T12:53:34.148-0800","caller":"cmd/restoreSnapshot.go:44","msg":"failed to restore snapshot","error":"snapshot hash file not found at 'https://eigenlayer-sidecar.s3.us-east-1.amazonaws.com/snapshots/testnet-holesky/sidecar-testnet-holesky_v3.0.0-rc.1_public_20250122.dump.sha256sum'. Ensure the file exists or set --verify-input=false to skip verification","stacktrace":"github.com/Layr-Labs/sidecar/cmd.init.func4\n\t/Users/brendan/Documents/GitHub/sidecar/cmd/restoreSnapshot.go:44\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/brendan/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/brendan/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/brendan/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041\ngithub.com/Layr-Labs/sidecar/cmd.Execute\n\t/Users/brendan/Documents/GitHub/sidecar/cmd/root.go:19\nmain.main\n\t/Users/brendan/Documents/GitHub/sidecar/main.go:8\nruntime.main\n\t/opt/homebrew/Cellar/go/1.23.4/libexec/src/runtime/proc.go:272"}

cmd/createSnapshot.go
```diff
@@ -37,8 +38,8 @@ var createSnapshotCmd = &cobra.Command{
 			return err
 		}
 
-		if err := svc.CreateSnapshot(); err != nil {
-			return fmt.Errorf("failed to create snapshot: %w", err)
+		if err := svc.RestoreSnapshot(); err != nil {
```
seanmcgary: why is `create-snapshot` now calling RestoreSnapshot?

cmd/createSnapshot.go
```diff
@@ -37,8 +38,8 @@ var createSnapshotCmd = &cobra.Command{
...
+			l.Fatal("failed to restore snapshot", zap.Error(err))
```
seanmcgary: `l.Sugar().Fatalw()`

cmd/restoreSnapshot.go
```diff
@@ -9,39 +9,39 @@ import (
...
+			DbName:      cfg.DatabaseConfig.DbName,
+			SchemaName:  cfg.DatabaseConfig.SchemaName,
 		}, l)
 		if err != nil {
 			return err
 		}
 
 		if err := svc.RestoreSnapshot(); err != nil {
-			return fmt.Errorf("failed to restore snapshot: %w", err)
+			l.Fatal("failed to restore snapshot", zap.Error(err))
```
seanmcgary:
```go
l.Sugar().Fatalw
```

pkg/snapshot/snapshot.go
```diff
@@ -1,52 +1,63 @@
...
+		cfg:       cfg,
+		l:         l,
+		tempFiles: []string{},
+	}, nil
+}
 
-	cfg.InputFile, err = resolveFilePath(cfg.InputFile)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve input file path: %w", err)
+// isNetworkURL checks if the given string is a network URL (http, https, ftp).
```
seanmcgary: Not doing ftp so that needs to be removed

pkg/snapshot/snapshot.go
```diff
@@ -1,52 +1,63 @@
...
+// Note: This function does not allow 'file' URLs as it's ambiguous with local file.
```
seanmcgary: We should clarify the function then and name it `isHttpURL()`

pkg/snapshot/snapshot.go
```diff
@@ -70,6 +81,14 @@ func resolveFilePath(path string) (string, error) {
 
 // CreateSnapshot creates a snapshot of the database based on the provided configuration.
 func (s *SnapshotService) CreateSnapshot() error {
+	resolvedFilePath, err := resolveFilePath(s.cfg.OutputFile)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to resolve output file path '%s'", s.cfg.OutputFile))
+	}
+	s.cfg.OutputFile = resolvedFilePath
```
seanmcgary: This is a bit of a bad practice. Config values should ideally not get changed and looking at the rest of this file, everything else properly takes the output file as an argument, which means you can just use `resolvedFilePath` in your function calls down below that currently use `s.cfg.OutputFile`

pkg/snapshot/snapshot.go
```diff
@@ -81,16 +100,34 @@ func (s *SnapshotService) CreateSnapshot() error {
...
+		s.l.Sugar().Errorw("Failed to create database snapshot",
+			zap.Error(dumpExec.Error.Err),
+			zap.String("output", dumpExec.Output),
+		)
 		return dumpExec.Error.Err
 	}
 
 	s.l.Sugar().Infow("Successfully created snapshot")
+
+	outputHashFile := getHashName(s.cfg.OutputFile)
```
seanmcgary: See above comment, this should be `getHashName(resolvedFilePath)`

pkg/snapshot/snapshot.go
```diff
@@ -81,16 +100,34 @@ func (s *SnapshotService) CreateSnapshot() error {
...
+	if err := saveOutputFileHash(s.cfg.OutputFile, outputHashFile); err != nil {
```
seanmcgary: Same with this one:

```go
saveOutputFileHash(resolvedFilePath, outputHashFile)
```

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +150,72 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
 }
 
+// resolveAndDownloadRestoreInput prepares the SnapshotService struct into a suitable format for restoring a snapshot
+// and downloads the necessary files if necessary
+func (s *SnapshotService) resolveAndDownloadRestoreInput() error {
+	if isNetworkURL(s.cfg.Input) {
+		inputUrl := s.cfg.Input
+
+		// Check if the input URL exists
+		snapshotExists, err := urlExists(inputUrl)
```
seanmcgary: why do we need to do this rather than just downloading the file? if the file doesnt exist, you'll get a 404 anyway so not sure why we need this extra request

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +150,72 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+		}
+		if !snapshotExists {
+			return fmt.Errorf("snapshot file not found at '%s'. Ensure the file exists", inputUrl)
+		}
+
+		// Check if the hash URL exists
+		if s.cfg.VerifyInput {
+			hashUrl := getHashName(inputUrl)
+
+			hashFileExists, err := urlExists(hashUrl)
```
seanmcgary: This I understand wanting to check since we wouldnt want to download multiple gigs of data and then have to re-do it because the corresponding hash doesnt exist, so this makes sense.

seanmcgary: Alternatively, since the hash file is super small, you could just download it rather than making a HEAD request and save yourself the request

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +150,72 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("error checking existence of snapshot hash URL '%s'", hashUrl))
+			}
+			if !hashFileExists {
+				return fmt.Errorf("snapshot hash file not found at '%s'. Ensure the file exists or set --verify-input=false to skip verification", hashUrl)
+			}
+		}
+
+		// Download the snapshot file and assign to s.cfg.Input
```
seanmcgary: except we're using `inputUrl` not `s.cfg.Input`. Comments should explain why, not what and/or clarify potentially complex code/edgecases

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +150,72 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+		fileName, err := getFileNameFromURL(inputUrl)
```
seanmcgary: Comment says we're downloading the file but this function is not downloading the file. Recommend removing the comment

pkg/snapshot/snapshot.go
```diff
@@ -113,6 +150,72 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
+			s.tempFiles = append(s.tempFiles, hashFile)
+
+		}
+	}
+
+	resolvedFilePath, err := resolveFilePath(s.cfg.Input)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to resolve input file path '%s'", s.cfg.Input))
+	}
+	s.cfg.Input = resolvedFilePath
```
seanmcgary: Same as above, we should not be overwriting config values

pkg/snapshot/snapshot.go
```diff
@@ -148,13 +251,24 @@ func (s *SnapshotService) setupSnapshotDump() (*pgcommands.Dump, error) {
...
 
 func (s *SnapshotService) validateRestoreConfig() error {
-	if s.cfg.InputFile == "" {
-		return fmt.Errorf("restore snapshot file path i.e. `input-file` must be specified")
+	if s.cfg.Input == "" {
+		return fmt.Errorf("restore snapshot file path i.e. `input` must be specified")
 	}
 
-	info, err := os.Stat(s.cfg.InputFile)
+	// s.cfg.input is resolved from a url in resolveAndDownloadRestoreInput()
```
seanmcgary: if we're having to leave comments because the meaning of our config values changes during runtime, we need to modify the behavior to prevent that. config values should be immutable in nature

pkg/snapshot/snapshot_test.go
seanmcgary: where? I dont see a test that shells out to `sha256sum`

seanmcgary (changes requested):

cmd/createSnapshot.go
BrendanCoughlan5: mb, switched to RestoreSnapshot()/ error

cmd/createSnapshot.go
BrendanCoughlan5: changed to l.Sugar().Fatal("failed to create snapshot", zap.Error(err))

cmd/restoreSnapshot.go
BrendanCoughlan5: switched to 		if err := svc.RestoreSnapshot(); err != nil {
			l.Sugar().Fatal("failed to restore snapshot", zap.Error(err))
		}

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed to 

// isHttpURL checks if the given string is a network URL (http, https)
func isHttpURL(str string) bool {

BrendanCoughlan5:
https://github.com/Layr-Labs/sidecar/pull/201#discussion_r1943190191

downloadFile is intended to be used to download an individual file, like you downloadFile(snapshot), then downloadFile(snapshotHash)

pkg/snapshot/snapshot_test.go
BrendanCoughlan5: it's not in this test, used TestSaveOutputFileHashCompatibilityWithSha256sum without shelling out, just against the contents of a .sha256sum, now made a bunch of the tests also shell out to sha256sum to be more robust

pkg/snapshot/snapshot.go
BrendanCoughlan5: now using resolvedFilePath

pkg/snapshot/snapshot.go
BrendanCoughlan5: now using resolvedFilePath

pkg/snapshot/snapshot.go
BrendanCoughlan5: now downloading instead of checking if exist then downloading

pkg/snapshot/snapshot.go
BrendanCoughlan5: now using resolvedFilePath

pkg/snapshot/snapshot.go
BrendanCoughlan5: now just downloading it

pkg/snapshot/snapshot.go
BrendanCoughlan5: updated to not change s.cfg.Input

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed comment

pkg/snapshot/snapshot.go
BrendanCoughlan5: now using resolvedFilePath and not overwriting config values

pkg/snapshot/snapshot.go
BrendanCoughlan5: now immutable config values

cmd/createSnapshot.go
```diff
@@ -38,7 +39,7 @@ var createSnapshotCmd = &cobra.Command{
 		}
 
 		if err := svc.CreateSnapshot(); err != nil {
-			return fmt.Errorf("failed to create snapshot: %w", err)
+			l.Sugar().Fatal("failed to create snapshot", zap.Error(err))
```
seanmcgary:
```go
l.Sugar().Fatalw()
```

cmd/restoreSnapshot.go
```diff
@@ -9,39 +9,39 @@ import (
...
+			DbName:      cfg.DatabaseConfig.DbName,
+			SchemaName:  cfg.DatabaseConfig.SchemaName,
 		}, l)
 		if err != nil {
 			return err
 		}
 
 		if err := svc.RestoreSnapshot(); err != nil {
-			return fmt.Errorf("failed to restore snapshot: %w", err)
+			l.Sugar().Fatal("failed to restore snapshot", zap.Error(err))
```
seanmcgary:
```bash
l.Sugar().Fatalw()
```

cmd/root.go
```diff
@@ -63,8 +63,10 @@ func init() {
 	rootCmd.AddCommand(rpcCmd)
 
 	// bind any subcommand flags
-	createSnapshotCmd.PersistentFlags().String(config.SnapshotOutputFile, "", "Path to save the snapshot file to (required)")
-	restoreSnapshotCmd.PersistentFlags().String(config.SnapshotInputFile, "", "Path to the snapshot file (required)")
+	createSnapshotCmd.PersistentFlags().String(config.SnapshotOutputFile, "", "Path to save the snapshot file to (required), also creates a hash file")
+
+	restoreSnapshotCmd.PersistentFlags().String(config.SnapshotInput, "", "Path to the snapshot file either a URL or a local file (required)")
```
seanmcgary: Awkward phrasing, recommend:

> Path to a local snapshot file, or remote URL to download from

cmd/root.go
```diff
@@ -63,8 +63,10 @@ func init() {
...
+	restoreSnapshotCmd.PersistentFlags().Bool(config.SnapshotVerifyInput, true, "Boolean to verify the input file against its .sha256sum file, if input is a url then it downloads the file, (default is true)")
```
seanmcgary: > Boolean to verify the input file against a sha256 checksum (`<file name>.sha256sum`). If a URL is provided, the checksum will be downloaded (default is true)

pkg/snapshot/snapshot.go
```diff
@@ -81,30 +98,110 @@ func (s *SnapshotService) CreateSnapshot() error {
...
+		}
+
+		// Download the snapshot file
+		err = downloadFile(inputUrl, inputFilePath)
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("error downloading snapshot from '%s'", inputUrl))
+		}
+		s.tempFiles = append(s.tempFiles, inputFilePath)
+
+		resolvedFilePath, err = resolveFilePath(inputFilePath)
```
seanmcgary: Shouldnt this already be resolved since we did `filepath.Join(os.TempDir(), fileName)` above?

seanmcgary: Additionally, if its not resolved, wouldnt we want to add the resolved path to the temp files list after resolving it and not before?

pkg/snapshot/snapshot.go
```diff
@@ -81,30 +98,110 @@ func (s *SnapshotService) CreateSnapshot() error {
...
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("failed to resolve input file path '%s'", s.cfg.Input))
+		}
+	}
+
+	s.l.Sugar().Debugw("Resolved file path", zap.String("resolvedFilePath", resolvedFilePath))
+
+	// validate snapshot against the hash file
+	if s.cfg.VerifyInput {
+		if err := validateInputFileHash(resolvedFilePath, getHashName(resolvedFilePath)); err != nil {
```
seanmcgary: we've called `getHashName` at least 3 times now in this function. Can we make this less repetitive?

seanmcgary: For this function specifically, since we assume the hash file name is the dump file name plus `.sha256sum` at the end, you could likely omit taking the hash name entirely

pkg/snapshot/snapshot.go
seanmcgary: Doesnt look like thats the case. Still calling `urlExists` which executes a HEAD request

pkg/snapshot/snapshot.go
seanmcgary: Still making a HEAD request

pkg/snapshot/snapshot_test.go
```diff
@@ -76,62 +84,176 @@ func TestSetupSnapshotDump(t *testing.T) {
...
+	assert.NoError(t, err, "Reading output hash file should not fail")
+
+	// Validate the format of the hash file
+	hashParts := strings.Fields(string(hashContent))
+	assert.Equal(t, 2, len(hashParts), "Output hash file should contain two parts: hash and filename")
+	assert.Equal(t, filepath.Base(outputFile), hashParts[1], "Output hash file should contain the correct filename")
+	assert.Equal(t, sha256.Size*2, len(hashParts[0]), "Output hash file should have the correct hash length")
+
+	// Compare the hash from sha256sum with the one generated by saveOutputFileHash
+	assert.Equal(t, sha256sumParts[0], hashParts[0], "Hashes should match between sha256sum and saveOutputFileHash")
```
seanmcgary: this all works, but the easier way is to just run

```
sha256sum -c <checksum file>
```

and check to make sure it exits successfully. This requires that the checksum file and the original file are in the same directory.

seanmcgary (changes requested):

pkg/snapshot/snapshot.go
BrendanCoughlan5: removed

cmd/root.go
BrendanCoughlan5: updated to Path to a local snapshot file, or remote URL to download from (required)

cmd/root.go
BrendanCoughlan5: updated Boolean to verify the input file against a sha256 checksum (<file name>.sha256sum). If a URL is provided, the checksum will be downloaded

cmd/restoreSnapshot.go
BrendanCoughlan5: updated

cmd/createSnapshot.go
BrendanCoughlan5: updated

pkg/snapshot/snapshot.go
BrendanCoughlan5: func validateInputFileHash(inputFile string) error {

    edit to be this

BrendanCoughlan5: i think we have to keep it at 3 though, one is for a url and two are for files

pkg/snapshot/snapshot_test.go
BrendanCoughlan5: changed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed naming to absoluteFilePath and  getAbsoluteFilePath() to be clearer

PR #216, opened 2025-02-03
feat: add StateRoot migration construct

seanmcgary:
This PR introduces the concept of a state migration. When introducing new features, it's possible that there is existing transaction data that was not codified into a model that now needs to be. Rather than truncating the database, re-indexing blocks resulting in modifying existing StateRoots, this migration construct allows us to populate new models with existing data, leaving existing StateRoots unchanged, and codifying the changes in a sub-tree whos root is represented in the top-level StateRoot tree.

Screenshot 2025-02-04 at 8 25 47 AM

The StateMigrator specifies which migrations run at which block height for a given chain. Multiple migrations can be run for a single block and will be run in the order specified. Each migration will produce a root representing the changes made. Each of those roots are composed of the same StateRoot structure used at the very top-level, one for each block that was touched in the migration process. In the majority of cases where we're populating models with only a few events, these StateRoot trees will be rather sparse and only include the new model.

Each migration works by:

* Creating a new instance of the StateManager
* Querying the database for the TransactionLogs that need to be processed
* Groups them by block number
* In the order of blockNumber ascending, feeds each log through the StateManager as it would have if it came through in real time.

The same lifecycle of HandleLogStateChange --> CommitFinalState --> GenerateStateRoot is followed.

## todo:

- [ ] Update fork block heights and dates


internal/config/config.go
```diff
@@ -288,36 +288,82 @@ func (c *Config) GetGenesisBlockNumber() uint64 {
...
+			},
+			RewardsFork_Arno: Fork{
+				Date: "2024-12-11",
+			},
+			RewardsFork_Trinity: Fork{
+				Date: "2025-01-09",
+			},
+			RewardsFork_Mississippi: Fork{
+				Date:        "2025-02-04",
+				BlockNumber: 3287006,
```
0xrajath, line 319: We need to update this with today's date and some block height later today right?

pkg/eigenState/stateManager/stateManager.go
```diff
@@ -109,7 +112,6 @@ func (e *EigenStateManager) CleanupProcessedStateForBlock(blockNumber uint64) er
 
 func (e *EigenStateManager) GenerateStateRoot(blockNumber uint64, blockHash string) (types.StateRoot, error) {
 	sortedIndexes := e.GetSortedModelIndexes()
-	common.FromHex(blockHash)
```
0xrajath, line 113: Was this deletion intended?

pkg/eigenState/stateMigrator/stateMigrator.go
```diff
@@ -0,0 +1,169 @@
...
+// getMigrations returns a map of block numbers to migrations that should be run at that block number
+// taking the chain into account
+func (sm *StateMigrator) initialize() error {
+	forks, err := sm.globalConfig.GetRewardsSqlForkDates()
+	if err != nil {
+		return err
+	}
+	migrations := NewMigrations()
+
+	if mississippiFork := forks[config.RewardsFork_Mississippi]; mississippiFork.BlockNumber != 0 {
```
0xrajath, line 69: Do you think it's better to check the chain directly vs assuming block number != 0 here?

0xrajath: i.e to be more explicit

internal/config/config.go
seanmcgary: Thats correct - Ive got it in a TODO in the PR description

pkg/eigenState/stateManager/stateManager.go
seanmcgary: Yeah because its not doing anything lol. A line or two below we make the same call but actually do something with the result, so no idea how this got left there

seanmcgary: Im actually really surprised the linter didnt catch this...

pkg/eigenState/stateMigrator/stateMigrations/202502031222_operatorSets/migration.go
```diff
@@ -0,0 +1,144 @@
...
+	contracts := sm.globalConfig.GetContractsMapForChain()
+
+	query := `
+		select
+			tl.*
+		from transaction_logs as tl
+		where
+			tl.address = @allocationManagerAddress
+			and event_name in @eventNames
+			and block_number < @currentBlockNumber
```
0xrajath, line 58: Just want to double check that we're indeed capturing state models in `currentBlockNumber` as part of the regular ingestion and not missing it.

pkg/eigenState/stateMigrator/stateMigrations/202502031222_operatorSets/migration.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, tl := range transactionLogs {
+		if _, ok := blocks[tl.BlockNumber]; !ok {
+			blocks[tl.BlockNumber] = make([]*storage.TransactionLog, 0)
+		}
+		blocks[tl.BlockNumber] = append(blocks[tl.BlockNumber], tl)
+	}
+
+	stateRoots := make([][]byte, 0)
+	committedStates := make(map[string][]interface{})
+	for bn, logs := range blocks {
```
0xrajath, line 89: To double check my understanding here, we're doing the entire state model pipeline just for these new events in all the old blocks right? (and not all the 'interesting' events in the old blocks).

pkg/eigenState/stateMigrator/stateMigrator.go
seanmcgary: I thought about a few options here. I settled on this one because checking each chain directly is going to get a bit messy.

What I wanted to do was check to see if `BlockNumber` is nil, but integer values cant be nil, so if a field is omitted from a struct, its automatically assigned the value of `0` by default. While not perfect, I think its fine because we'll never hit a block height of `0` so in this case it's our substitution for a null value

0xrajath (dismissed):
LGTM! I think we just need to update the fork height and date for preprod.

pkg/eigenState/stateMigrator/stateMigrations/202502031222_operatorSets/migration.go
seanmcgary: Just the StateManager lifecycle. Here, we take all of our logs, batch them by block number and for each blockNumber, run the StateManager like it would be if we were processing a block in the main index loop

0xrajath (dismissed):

0xrajath (approved):

PR #226, opened 2025-02-10
feat: add GetRewardsByAvsForDistributionRoot rpc

### LINKED ISSUE #211: feat: add RPC to get earned rewards, segmented by AVS

seanmcgary:
closes #211

### DIFF
--- a/pkg/rpcServer/rewardsHandlers.go
+++ b/pkg/rpcServer/rewardsHandlers.go
@@ -430,6 +430,9 @@ func convertRewardTypeToEnum(rewardType string) (rewardsV1.RewardType, error) {
 	}
 }
 
+// GetRewardsByAvsForDistributionRoot returns the rewards for a specific distribution root.
+//
+// TODO(seanmcgary): add pagination if this response gets too large in the future
 func (rpc *RpcServer) GetRewardsByAvsForDistributionRoot(ctx context.Context, req *rewardsV1.GetRewardsByAvsForDistributionRootRequest) (*rewardsV1.GetRewardsByAvsForDistributionRootResponse, error) {
 	rootIndex := req.GetRootIndex()
 
--- a/pkg/rpcServer/rewardsHandlers.go
+++ b/pkg/rpcServer/rewardsHandlers.go
@@ -3,6 +3,7 @@ package rpcServer
 import (
 	"context"
 	"errors"
+	"fmt"
 	rewardsV1 "github.com/Layr-Labs/protocol-apis/gen/protos/eigenlayer/sidecar/v1/rewards"
 	"github.com/Layr-Labs/sidecar/pkg/metaState/types"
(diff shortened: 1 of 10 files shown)


pkg/rpcServer/rewardsHandlers.go
```diff
@@ -416,6 +417,45 @@ func (rpc *RpcServer) ListDistributionRoots(ctx context.Context, req *rewardsV1.
...
+	case "all_earners":
+		return rewardsV1.RewardType_REWARD_TYPE_FOR_ALL_EARNERS, nil
+	default:
+		return -1, fmt.Errorf("unknown reward type '%s'", rewardType)
+	}
+}
+
 func (rpc *RpcServer) GetRewardsByAvsForDistributionRoot(ctx context.Context, req *rewardsV1.GetRewardsByAvsForDistributionRootRequest) (*rewardsV1.GetRewardsByAvsForDistributionRootResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "method GetRewardsByAvsForDistributionRoot not implemented")
+	rootIndex := req.GetRootIndex()
```
anupsv, line 437: Does this need to be constrained ? I.e, the index cannot be more than a certain number, limit etc ?

pkg/rpcServer/rewardsHandlers.go
```diff
@@ -416,6 +417,45 @@ func (rpc *RpcServer) ListDistributionRoots(ctx context.Context, req *rewardsV1.
...
+
+	rewards, err := rpc.rewardsDataService.GetRewardsByAvsForDistributionRoot(ctx, rootIndex)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	rewardsResponse := make([]*rewardsV1.AvsReward, 0, len(rewards))
```
anupsv, line 444: I assume this is done like this for now until pagination or something similar can be implemented?

pkg/rpcServer/rewardsHandlers.go
seanmcgary: As long as it fits in a `uint64`, its technically valid (which is enforced by the protobuf definition).

pkg/rpcServer/rewardsHandlers.go
seanmcgary: Thats correct. Right now it just returns everything since that's the expected use case. This route isnt currently intended to be open to the public via the platform and is to be used by anyone self-hosting the sidecar for their own use.

If we were to open it via the platform, we'd want to add proper pagination.

anupsv: Sounds good. Could you add a comment or something so that we remember for future.

seanmcgary: added

PR #229, opened 2025-02-11
chore: codeql compatible go.mod

anupsv:
Updating the go.mod file to match the specs needed for codeql build.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/Layr-Labs/sidecar
 
-go 1.23
+go 1.23.6
+
+toolchain go1.23.6
 
 require (
 	github.com/DataDog/datadog-go/v5 v5.5.0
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/Layr-Labs/sidecar
 
-go 1.23
+go 1.23.6
+
+toolchain go1.23.6
```
seanmcgary: because the go version in the above line is specific down to the patch release, `toolchain` is actually redundant. If `go` was left at `1.23` then toolchain would be appropriate (go mod actually removes the toolchain line for you automatically in this case...)

seanmcgary (approved):

go.mod
anupsv: Interesting, codeql was complaining for both. Maybe their parser needs updating.

PR #237, opened 2025-02-11
feat(sidecar-extensibility): detect upgradeable contracts and pull ABIs from IPFS

### LINKED ISSUE #223: feat: dynamically upgrade side-loaded contract implementations

---

serichoi65:
## Description

In order to side-load contracts to Sidecar, we want to add the following features:
- Detecting an upgradeable contract when we process transaction logs
  - This will not be applicable to EigenLayers' core contracts but only to side-loaded contracts to automatically update contract ABIs
  - This function is **not** being called in this PR 
 - Pulling ABIs from IPFS (main source)
    - The contract metadata needs to exist in IPFS in order for us to pull ABIs. Added 10 seconds timeout during HTTP Get
    - TODO: add a fallback source to pull the data from Etherscan

Non-feature updates:
- Added `AbiFetcher` as a new input to `ContractManager`
- All tests, except `pipelineIntegration_test` and `restakedStrategies_test`, should use a non-paid quiknode url address
- Updated the ethereum `NewClient` parameter with `http.Client` to accept mocked client

## Type of change

## How Has This Been Tested?

Added tests to prove my fix is effective for all new modules by running 
```
./scripts/goTest.sh -timeout 0 ./pkg/contractManager -v -p 1 -run 'Test_ContractManager'
```

## Checklist:


pkg/indexer/indexer.go
```diff
@@ -156,6 +157,55 @@ func (idx *Indexer) ParseInterestingTransactionsAndLogs(ctx context.Context, fet
...
+
+		newProxiedAddress = storageValue[26:]
+	}
+
+	if newProxiedAddress == "" {
+		idx.Logger.Sugar().Debugw("No new proxied address found", zap.String("address", upgradedLog.Address))
+		return nil
+	}
+
+	_, _, err := idx.ContractStore.FindOrCreateProxyContract(blockNumber, upgradedLog.Address, newProxiedAddress)
```
serichoi65, line 199: In block-lake, we used to use function `idx.ContractManager.CreateProxyContract` [code](https://github.com/Layr-Labs/block-lake/blob/master/internal/contractManager/contractManager.go#L208). Is this a safe replacement?

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,92 @@
...
+	mds := pgStorage.NewPostgresBlockStore(grm, l, cfg)
+
+	fetchr := fetcher.NewFetcher(client, cfg, l)
+
+	mccc := multicallContractCaller.NewMulticallContractCaller(client, l)
+
+	cm := contractManager.NewContractManager(contractStore, client, sdc, l)
+
+	t.Run("Test indexing contract upgrades", func(t *testing.T) {
+		contractAddress := "0x055733000064333caddbc92763c58bf0192ffebf"
```
serichoi65, line 60: Using this example for test:
```
blocklake=> select * from proxy_contracts where contract_address='0x055733000064333caddbc92763c58bf0192ffebf';
 block_number |              contract_address              |           proxy_contract_address           |         created_at         |          updated_at           |       deleted_at       
--------------+--------------------------------------------+--------------------------------------------+----------------------------+-------------------------------+------------------------
      1167044 | 0x055733000064333caddbc92763c58bf0192ffebf | 0xef5ba995bc7722fd1e163edf8dc09375de3d3e3a | 2024-06-05 00:14:04.013855 | 2024-06-05 00:14:04.013855+00 | 0001-01-01 00:00:00+00
      2969816 | 0x055733000064333caddbc92763c58bf0192ffebf | 0xa504276dfdee6210c26c55385d1f793bf52089a0 | 2024-12-19 18:37:01.865369 | 2024-12-19 18:37:01.865369+00 | 0001-01-01 00:00:00+00
      3091933 | 0x055733000064333caddbc92763c58bf0192ffebf | 0x69660e721e4013dd8fef83dce731e915d74b8a0b | 2025-01-06 20:59:13.897874 | 2025-01-06 20:59:13.897874+00 | 0001-01-01 00:00:00+00
(3 rows)
```

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,92 @@
...
+                    Value: common.HexToAddress("0xa504276dfdee6210c26c55385d1f793bf52089a0"),
+                    Indexed: true,
+                },
+            },
+        }
+
+		idxr := NewIndexer(mds, contractStore, cm, client, fetchr, mccc, grm, l, cfg)
+
+
+        // Get initial state
```
serichoi65, line 80: What would be the best way to make sure the contract upgrade worked? the number of rows in `contract_address` table?

pkg/contractManager/contractManager.go
```diff
@@ -43,3 +44,66 @@ func (cm *ContractManager) GetContractWithProxy(
...
+				zap.String("contractAddress", contractAddress),
+				zap.String("proxyContractAddress", proxyContractAddress),
+			)
+		} else {
+			cm.Logger.Sugar().Debugw("Created proxy contract",
+				zap.String("contractAddress", contractAddress),
+				zap.String("proxyContractAddress", proxyContractAddress),
+			)
+		}
+	}
```
seanmcgary, line 131: I'd simplify this a bit:

```go
if err != nil {
    cm.Logger.Sugar().Errorw("Failed to create proxy contract",
      zap.Error(err),
      zap.String("contractAddress", contractAddress),
      zap.String("proxyContractAddress", proxyContractAddress),
    )
    return nil, err
}
if found {
  cm.Logger.Sugar().Debugw("Found existing proxy contract",
    zap.String("contractAddress", contractAddress),
    zap.String("proxyContractAddress", proxyContractAddress),
  )
  return proxyContract, nil
}

cm.Logger.Sugar().Debugw("Created proxy contract",
  zap.String("contractAddress", contractAddress),
  zap.String("proxyContractAddress", proxyContractAddress),
)
```

That said, I would actually avoid using the`FindOrCreateProxyContract` at all and break this into two calls:

1. Get the proxy contract by implementation address and its proxy. If it exists, return it
2. Explicitly create it once you have all of the data to do so

In blocklake we did it the way you have it here because we had the ability to backfill and retry. In the sidecar, we dont have that luxury, so rather than upsert the proxy with only half the data, then update it with more info, but then potentially deal with having to purge it later since it can only half exist, we want to make a single atomic insert.

pkg/contractManager/contractManager.go
```diff
@@ -43,3 +44,66 @@ func (cm *ContractManager) GetContractWithProxy(
...
+	
+	bytecode, err := cm.EthereumClient.GetCode(context.Background(), proxyContractAddress)
```
seanmcgary: If we need to take a context object, we should have `CreateProxyContract` take it as an argument and pass it along

pkg/contractManager/contractManager.go
```diff
@@ -43,3 +44,66 @@ func (cm *ContractManager) GetContractWithProxy(
...
+		return nil, err
+	}
+
+	bytecodeHash := ethereum.HashBytecode(bytecode)
+	cm.Logger.Sugar().Debug("Fetched proxy contract bytecode",
+		zap.String("proxyContractAddress", proxyContractAddress),
+		zap.String("bytecodeHash", bytecodeHash),
+	)
+
+	_, _, err = cm.ContractStore.FindOrCreateContract(
```
seanmcgary: Like I said above, we should refactor this whole create function to create the entire proxy contract in one call rather than upserting it, fetching data, then updating it.

pkg/fetcher/fetcher.go
```diff
@@ -235,3 +236,14 @@ func (f *Fetcher) FetchBlocks(ctx context.Context, startBlockInclusive uint64, e
 
 	return fetchedBlocks, nil
 }
+
+func (f *Fetcher) GetContractStorageSlot(ctx context.Context, contractAddress string, blockNumber uint64) (string, error) {
+	stringBlock := ""
+	if blockNumber == 0 {
```
seanmcgary: Since we process everything block by block, we want this to explicitly be the block height we're processing, i.e. the provided `blockHeight` argument

seanmcgary: If it's 0, we should return an error

pkg/indexer/indexer.go
```diff
@@ -156,6 +157,55 @@ func (idx *Indexer) ParseInterestingTransactionsAndLogs(ctx context.Context, fet
 	return parsedTransactions, nil
 }
 
+func (idx *Indexer) IndexContractUpgrade(ctx context.Context, blockNumber uint64, upgradedLog *parser.DecodedLog) error {
```
seanmcgary: Rather than this living in the `Indexer` this should probably be in the contract manager and be called something like `HandleContractUpgrade`

pkg/indexer/indexer.go
```diff
@@ -156,6 +157,55 @@ func (idx *Indexer) ParseInterestingTransactionsAndLogs(ctx context.Context, fet
...
+	// the new address that the contract points to
+	newProxiedAddress := ""
+
+	// Check the arguments for the new address. EIP-1967 contracts include this as an argument.
+	// Otherwise, we'll check the storage slot
+	for _, arg := range upgradedLog.Arguments {
+		if arg.Name == "implementation" && arg.Value != "" && arg.Value != nil {
+			newProxiedAddress = arg.Value.(common.Address).String()
+		}
+	}
```
seanmcgary: immediately after this loop, I would short-circuit:

```
if newProxiedAddress != "" {
    // create it and return
}
```

pkg/indexer/indexer.go
```diff
@@ -156,6 +157,55 @@ func (idx *Indexer) ParseInterestingTransactionsAndLogs(ctx context.Context, fet
...
 
+func (idx *Indexer) IndexContractUpgrade(ctx context.Context, blockNumber uint64, upgradedLog *parser.DecodedLog) error {
+	// the new address that the contract points to
+	newProxiedAddress := ""
+
+	// Check the arguments for the new address. EIP-1967 contracts include this as an argument.
+	// Otherwise, we'll check the storage slot
+	for _, arg := range upgradedLog.Arguments {
+		if arg.Name == "implementation" && arg.Value != "" && arg.Value != nil {
+			newProxiedAddress = arg.Value.(common.Address).String()
```
seanmcgary: Once we've found the address, we can break out of the loop

pkg/indexer/indexer.go
```diff
@@ -156,6 +157,55 @@ func (idx *Indexer) ParseInterestingTransactionsAndLogs(ctx context.Context, fet
...
+			)
+			return err
+		}
+
+		newProxiedAddress = storageValue[26:]
+	}
+
+	if newProxiedAddress == "" {
+		idx.Logger.Sugar().Debugw("No new proxied address found", zap.String("address", upgradedLog.Address))
+		return nil
```
seanmcgary: If we're running this function with the assumption that we encountered an "Upgraded" event, this should probably be treated as an error case.

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+func Test_Indexer(t *testing.T) {
+	// uses setup from restakedStrategies_test
+	dbName, grm, l, cfg, err := setup()
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	baseUrl := "https://winter-white-crater.ethereum-holesky.quiknode.pro/1b1d75c4ada73b7ad98e1488880649d4ea637733/"
```
seanmcgary: We should probably make this a node that _isnt_ one of our paid quiknode endpoints

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	fetchr := fetcher.NewFetcher(client, cfg, l)
+
+	mccc := multicallContractCaller.NewMulticallContractCaller(client, l)
+
+	cm := contractManager.NewContractManager(contractStore, client, sdc, l)
+
+	t.Run("Test indexing contract upgrades", func(t *testing.T) {
+		// Create a contract
+		_, found, err := contractStore.FindOrCreateContract(contract.ContractAddress, contract.ContractAbi, contract.Verified, contract.BytecodeHash, contract.MatchingContractAddress, false)
```
seanmcgary: This is an okay way to do this. IMO I would create the models directly in the db with `grm.Model().Create()`, main reason being it saves you (and anyone else) from needing to update this later if `FindOrCreateContract` changes for some reason.

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,135 @@
...
+		// An upgrade event
+		upgradedLog := &parser.DecodedLog{
+			LogIndex:  0,
+			Address:   contract.ContractAddress,
+			EventName: "Upgraded",
+			Arguments: []parser.Argument{
+				{
+					Name:    "implementation",
+					Type:    "address",
+					Value:   common.HexToAddress("0x789"),
```
seanmcgary: Would recommend using a valid address here just to make it as complete as possible

pkg/indexer/indexer_test.go
```diff
@@ -0,0 +1,135 @@
...
+		// Verify database state after upgrade
+		newProxyContractAddress := upgradedLog.Arguments[0].Value.(common.Address).Hex()
+		res = grm.Raw(`select count(*) from contracts where contract_address=@newProxyContractAddress`, sql.Named("newProxyContractAddress", newProxyContractAddress)).Scan(&contractCount)
+		assert.Nil(t, res.Error)
+		assert.Equal(t, 1, contractCount)
+
+		res = grm.Raw(`select count(*) from proxy_contracts where contract_address=@contractAddress`, sql.Named("contractAddress", contractAddress)).Scan(&proxyContractCount)		
+		assert.Nil(t, res.Error)
+		assert.Equal(t, 2, proxyContractCount)
+	})
```
seanmcgary: Probably need a test as well to cover getting the address from the storage slot

pkg/pipeline/pipeline.go
```diff
@@ -181,6 +181,10 @@ func (p *Pipeline) RunForFetchedBlock(ctx context.Context, block *fetcher.Fetche
 				)
 				return err
 			}
+
+			if log.EventName == "Upgraded" {
+				p.Indexer.IndexContractUpgrade(ctx, blockNumber, log)
```
seanmcgary: Returned error needs to be handled correctly

seanmcgary (changes requested):

non-fungible-nelson:
Hi seri - great PR! Can we have a description please?

seanmcgary:
@serichoi65 I'll let you fill in the PR body with implementation specifics.

@non-fungible-nelson I added a link to the related github issue this addresses

pkg/fetcher/fetcher.go
serichoi65: It was used 0 in block-lake, so irrelevant here - removing.

cmd/database.go
```diff
@@ -6,6 +6,7 @@ import (
 	"github.com/Layr-Labs/sidecar/internal/logger"
 	"github.com/Layr-Labs/sidecar/internal/metrics"
 	"github.com/Layr-Labs/sidecar/internal/version"
+    "github.com/Layr-Labs/sidecar/pkg/abiFetcher"
```
seanmcgary: Wonky formatting. If you dont already have it running, recommend setting up `gofmt` to run automatically

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+    // Skip the 1220 prefix when decoding
+    bytes, err := hex.DecodeString(ipfsHash)
+    if err != nil {
+        return "", fmt.Errorf("failed to decode IPFS hash: %v", err)
+    }
+    
+    // Convert to base58
+    base58Hash := base58.Encode(bytes)
+    
+    return fmt.Sprintf("https://ipfs.io/ipfs/%s", base58Hash), nil
```
seanmcgary: This should definitely be taken as a config value that can be overridden at runtime.

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+			zap.Error(err),
+			zap.String("address", address),
+		)
+		return "", err
+	}
+	af.Logger.Sugar().Debug("Successfully retrieved IPFS URL",
+		zap.String("address", address),
+		zap.String("IPFS URL", url),
+	)
+	fmt.Printf("IPFS URL: %s \n", url)
```
seanmcgary: Let's remove this

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		af.Logger.Sugar().Errorw("Failed to get IPFS URL from bytecode",
+			zap.Error(err),
+			zap.String("address", address),
+		)
+		return "", err
+	}
+	af.Logger.Sugar().Debug("Successfully retrieved IPFS URL",
+		zap.String("address", address),
+		zap.String("IPFS URL", url),
```
seanmcgary: A bit of a style thing, but the `key` argument should be `camelCase` since log logs get encoded as json for easy machine parsing

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+	)
+	fmt.Printf("IPFS URL: %s \n", url)
+
+    httpClient := &http.Client{
```
seanmcgary: To make things testable, an http client reference should probably be passed to the NewAbiFetcher. That way you could pass in a mocked client.

pkg/abiFetcher/abiFetcher_test.go
```diff
@@ -0,0 +1,79 @@
...
+	client := ethereum.NewClient(ethConfig, l)
+
+	t.Run("Test getting IPFS url from bytecode", func(t *testing.T) {
+		af := NewAbiFetcher(client, l)
+
+		address := "0x29a954e9e7f12936db89b183ecdf879fbbb99f14"
+		bytecode, err := af.EthereumClient.GetCode(context.Background(), address)
+		assert.Nil(t, err)
+
+		url, err := af.GetIPFSUrlFromBytecode(bytecode)
```
seanmcgary, line 70: If possible, I would at least mock the IPFS http call. What we dont want is to rely on a live service that we dont control that might break and cause our tests to fail unexpectedly.

pkg/clients/ethereum/client.go
```diff
@@ -96,10 +96,13 @@ func DefaultChunkedCallEthereumClientConfig() *EthereumClientConfig {
 	}
 }
 
-func NewClient(cfg *EthereumClientConfig, l *zap.Logger) *Client {
+func NewClient(cfg *EthereumClientConfig, l *zap.Logger, hc ...*http.Client) *Client {
 	client := &http.Client{
 		Timeout: time.Second * 10,
 	}
+	if len(hc) > 0 && hc[0] != nil {
+        client = hc[0]
```
seanmcgary: Better to make the argument `httpClient *http.Client` and just check if its nil.

As-is, yeah you can omit the argument all together, but it also is misleading because someone calling `NewClient` may think it can take multiple clients when really it only uses one.

Another alternative solution (which is probably cleaner) is to not take a client at all in this function but add a `SetHttpClient(c *http.Client)` function to override the default.

pkg/contractManager/contractManager.go
```diff
@@ -43,3 +52,110 @@ func (cm *ContractManager) GetContractWithProxy(
 
 	return contract, nil
 }
+
+// this function is called when there's an "Upgraded" event
```
seanmcgary: Go function comments should start with the function name and describe what it does:

```
// HandleContractUpgrade parses an Upgraded contract log and inserts the new upgraded implementation into the database
```

pkg/contractManager/contractManager.go
```diff
@@ -43,3 +52,110 @@ func (cm *ContractManager) GetContractWithProxy(
...
+		if len(storageValue) != 66 {
+			cm.Logger.Sugar().Errorw("Invalid storage value",
+				zap.Uint64("block", blockNumber),
+				zap.String("storageValue", storageValue),
+			)
+			return err
+		}
+
+		newProxiedAddress = "0x" + storageValue[26:]
+		if newProxiedAddress == "" {
```
seanmcgary: I would move this `if` block outside of the current `if newProxiedAddress == ""` block its in. This way we for sure can catch a potential empty string

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+		)
+		return "", "", err
+	}
+
+	bytecodeHash := ethereum.HashBytecode(bytecode)
+	af.Logger.Sugar().Debug("Fetched the contract bytecodeHash",
+		zap.String("address", address),
+		zap.String("bytecodeHash", bytecodeHash),
+	)
+	fmt.Printf("bytecodeHash: %s", bytecodeHash)
```
seanmcgary: this can be removed

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+	e *ethereum.Client,
+	l *zap.Logger,
+) *AbiFetcher {
+	return &AbiFetcher{
+		EthereumClient: e,
+		Logger:         l,
+	}
+}
+
+func (af *AbiFetcher) FetchMetadataFromAddress(ctx context.Context, address string) (string, string, error) {
```
seanmcgary: Naming is a little funky since this returns the ABI not the metadata

pkg/abiFetcher/abiFetcher.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	bytecodeHash := ethereum.HashBytecode(bytecode)
+	af.Logger.Sugar().Debug("Fetched the contract bytecodeHash",
+		zap.String("address", address),
+		zap.String("bytecodeHash", bytecodeHash),
+	)
+	fmt.Printf("bytecodeHash: %s", bytecodeHash)
+
+	// fetch ABI using IPFS
```
seanmcgary, line 61: Since we want to add Etherscan later (or even other options), I would refactor this a little bit into two components:

```go
type AbiFetcher interface {
	GetContractAbiAndMetadata(ctx context.Context, address string) (string, error)
}

type IAbiSource interface {
	FetchAbi(ctx context.Context, address string, bytecode string) (string, error)
}
```

Move all of the IPFS logic into an implementation of `IAbiSource` and have the current `AbiFetcher` take a list of `IAbiSource`s when created. Then all you need to do is iterate over the list in order until you find an ABI or error out.

seanmcgary (changes requested):

pkg/abiFetcher/abiFetcher.go
serichoi65: It returns `bytecodeHash` and `ABI` that I called it metadata, but do you prefer it to be `FetchABIFromAddress`?

pkg/contractManager/contractManager.go
serichoi65: That was my personal comment that should've been removed, but I think it's a good idea to add your suggestion!

pkg/abiFetcher/abiFetcher.go
serichoi65: I'll refactor this in the PR when I add Etherscan

seanmcgary (approved):

PR #245, opened 2025-02-12
feat(restore-snapshot): add manifest URL flags

BrendanCoughlan5:
add the manifest


seanmcgary:
FYI PRs should always target `master` or a corresponding release branch, not another feature branch.

pkg/snapshot/snapshot.go
```diff
@@ -122,73 +126,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
-		}
-		if !snapshotExists {
-			return errors.Wrap(fmt.Errorf("snapshot file not found at '%s'. Ensure the file exists", inputUrl), "snapshot file not found")
-		}
-
-		fileName, err := getFileNameFromURL(inputUrl)
+	var snapshotAbsoluteFilePath string
+	var err error
+	if s.cfg.Input == "" && s.cfg.ManifestURL == "" {
+		return errors.Wrap(fmt.Errorf("input file or manifest URL is required"), "missing required configuration")
```
seanmcgary: Theres no error to wrap, so this should just be `fmt.Errorf`

pkg/snapshot/snapshot.go
```diff
@@ -122,73 +126,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
-		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("error checking existence of snapshot URL '%s'", inputUrl))
-		}
-		if !snapshotExists {
-			return errors.Wrap(fmt.Errorf("snapshot file not found at '%s'. Ensure the file exists", inputUrl), "snapshot file not found")
-		}
-
-		fileName, err := getFileNameFromURL(inputUrl)
+	var snapshotAbsoluteFilePath string
+	var err error
```
seanmcgary: no need to pre-declare the `err` variable

pkg/snapshot/snapshot.go
```diff
@@ -122,73 +126,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
-				return errors.Wrap(err, fmt.Sprintf("error downloading snapshot hash from '%s'", hashUrl))
-			}
-			s.tempFiles = append(s.tempFiles, hashFilePath)
+			return err
 		}
-
-		// Download the snapshot file
-		err = downloadFile(inputUrl, inputFilePath)
+		// Use the desired URL as the input for downloading
+		snapshotAbsoluteFilePath, err = s.downloadSnapshotAndVerificationFiles(desiredURL)
```
seanmcgary: this can be more DRY where you only need to call the download function once. See above comment

pkg/snapshot/snapshot.go
```diff
@@ -122,73 +126,45 @@ func (s *SnapshotService) RestoreSnapshot() error {
...
-		if !snapshotExists {
-			return errors.Wrap(fmt.Errorf("snapshot file not found at '%s'. Ensure the file exists", inputUrl), "snapshot file not found")
-		}
-
-		fileName, err := getFileNameFromURL(inputUrl)
+	var snapshotAbsoluteFilePath string
+	var err error
+	if s.cfg.Input == "" && s.cfg.ManifestURL == "" {
+		return errors.Wrap(fmt.Errorf("input file or manifest URL is required"), "missing required configuration")
+	} else if s.cfg.Input == "" && s.cfg.ManifestURL != "" {
```
seanmcgary: This if-else chain is kinda confusing. Would restructure it as:

```go
input := s.cfg.Input
if input == "" && s.cfg.ManifestURL == "" {
  // return error
}

if input == "" {
  // load snapshot data from manifest, overwriting input with the url from the manifest

}

if isHttpURL(input) {
  // handle downloading
}
```

pkg/snapshot/snapshot.go
```diff
@@ -389,22 +369,117 @@ func getFileNameFromURL(rawURL string) (string, error) {
...
+	MetadataVersion string `json:"metadataVersion"`
+	LastRefreshed   string `json:"lastRefreshed"`
+}
+
+type Snapshot struct {
+	SnapshotURL string `json:"snapshotURL"`
+	Chain       string `json:"chain"`
+	Version     string `json:"version"`
+	Schema      string `json:"schema"`
+	Timestamp   string `json:"timestamp"`
```
seanmcgary: Should be parsed to a `time.Time` type

pkg/snapshot/snapshot.go
```diff
@@ -389,22 +369,117 @@ func getFileNameFromURL(rawURL string) (string, error) {
...
+}
+
+func (s *SnapshotService) getDesiredURLFromManifest(manifestURL string) (string, error) {
+	// Fetch the snapshot URL from the manifest
+	if !isHttpURL(manifestURL) {
+		return "", errors.Wrap(fmt.Errorf("manifest URL must be a network URL"), "invalid manifest URL")
+	}
+
+	// Download the manifest file
+	manifestFileName, err := getFileNameFromURL(manifestURL)
```
seanmcgary: Why do we need to download the manifest file to disk?

pkg/snapshot/snapshot_test.go
```diff
@@ -341,23 +342,191 @@ func TestCreateAndRestoreSnapshot(t *testing.T) {
 		err = svc.RestoreSnapshot()
 		assert.NoError(t, err, "Restoring snapshot should not fail")
 
-		// Validate the restore process
+		// Step 1: Validate the restore process
+		var countBefore int64
+		dbGrm.Raw("SELECT COUNT(*) FROM migrations").Scan(&countBefore)
```
seanmcgary: With an empty database, `migrations` shouldnt even exist so this query would fail

seanmcgary (changes requested):

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed

pkg/snapshot/snapshot.go
BrendanCoughlan5: changed to fetch instead

pkg/snapshot/snapshot.go
BrendanCoughlan5: alternatively do it this way :)

	var err error
		input, err = s.getDesiredURLFromManifest(s.cfg.ManifestURL)
		if err != nil {
			return err
		}

pkg/snapshot/snapshot_test.go
BrendanCoughlan5: catches the empty db case if the restore failed.

PR #265, opened 2025-02-25
feat(sidecar-extensibiliity): add Etherscan to pull ABIs

### LINKED ISSUE #223: feat: dynamically upgrade side-loaded contract implementations

serichoi65:
## Description

This PR creates an abstract `AbiSource` interface to accept multiple ABI source implementations and allows Etherscan to pull ABIs as one of the methods.
- Part of `abiFetcher.go` code that pulled ABIs from IPFS has been refactored to `ipfsAbiSource.go`
- Added `etherscanAbiSource.go`
- `abiFetcher` now accepts a list of sources to go through to pull ABIs

## Type of change

## How Has This Been Tested?

Unit tests

## Checklist:


internal/config/config.go
```diff
@@ -187,6 +193,9 @@ var (
 	SidecarPrimaryUrl = "sidecar-primary.url"
 
 	IpfsUrl = "ipfs.url"
+
+	EtherscanUrl    = "etherscan.url"
+	EtherscanApiKey = "etherscan.api_key"
```
seanmcgary: Can we use a dash here rather than an underscore? Somehow we started using underscores and Im not sure how/why...

internal/config/config.go
```diff
@@ -253,6 +262,11 @@ func NewConfig() *Config {
 		IpfsConfig: IpfsConfig{
 			Url: StringWithDefault(viper.GetString(normalizeFlagName(IpfsUrl)), "https://ipfs.io/ipfs"),
 		},
+
+		EtherscanConfig: EtherscanConfig{
+			Url:    StringWithDefault(viper.GetString(normalizeFlagName(EtherscanUrl)), "https://%s.etherscan.io/api?"),
```
seanmcgary: Since the URL for the Etherscan API is well known, we probably shouldnt take it at all and just take an API key since I cant think of a case that you'd ever override the default

pkg/abiFetcher/abiFetcher.go
```diff
@@ -58,101 +51,12 @@ func (af *AbiFetcher) FetchContractDetails(ctx context.Context, address string)
...
-			zap.Error(err),
-			zap.String("address", address),
-		)
-		return "", err
+	// fetch ABI in order of AbiSource implementations
+	for _, source := range af.AbiSources {
+		abi, err := source.FetchAbi(address, bytecode)
+		if err == nil {
+			return bytecodeHash, abi, nil
+		}
```
seanmcgary, line 62: Probably worth logging a warning here

pkg/abiSource/abiSource.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"encoding/json"
+)
+
+type AbiSource interface {
+	FetchAbi(address string, bytecode string) (string, error)
+}
+
+type Response struct {
```
seanmcgary: Since this is specific to the IPFS source, it should be moved there

pkg/abiSource/etherscanAbiSource/etherscanAbiSource.go
```diff
@@ -0,0 +1,35 @@
+package etherscanAbiSource
```
seanmcgary: Im lazy, so Id just shorten this to `etherscan` and avoid all the typing, since abiSource is the parent package 😛

pkg/abiSource/etherscanAbiSource/etherscanAbiSource.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/clients/etherscan"
+	"go.uber.org/zap"
+)
+
+type EtherscanAbiSource struct {
+	EtherscanClient *etherscan.EtherscanClient
+	Logger          *zap.Logger
```
seanmcgary: This is something Ive been meaning to add to the linter, but both of these can be lower-cased and private to the struct. In a lot of places we (erroneously) expose struct members when we shouldnt be

pkg/abiSource/ipfsAbiSource/ipfsAbiSource.go
```diff
@@ -0,0 +1,115 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/pkg/abiSource"
+	"github.com/btcsuite/btcutil/base58"
+	"go.uber.org/zap"
+)
+
+type IpfsAbiSource struct {
```
seanmcgary: Same here - since `abiSource` is the parent package, we can name this `Ipfs`

internal/config/config.go
seanmcgary: Also, make sure to add the flags in `cmd/root.go` so they'll actually get loaded into viper.

seanmcgary (changes requested):

pkg/abiSource/etherscanAbiSource/etherscanAbiSource.go
serichoi65: Updated `Ipfs` struct to be lower-cased too.

internal/config/config.go
serichoi65: I'll add the flags when I have the next PR, working on the cli to connect all parts together!

seanmcgary (approved):

PR #273, opened 2025-02-27
feat(sidecar-extensibility): add load-contract cli to side-load any contracts to sidecar

### LINKED ISSUE #191: feat: side-load additional contracts for indexing
## Summary

Currently the Sidecar is responsible for indexing only the core contracts that make up the EigenLayer protocol. We've heard from other potential customers (AVSs, operators, LRTs, custodians, etc) that their existing use-cases require data from additional contracts.

The goal of this feature is to give customers the option of being able to index the logs and events from any other contracts they desire, without polluting the default state of the Sidecar which will remain focused on the EigenLayer protocol.

## Requirements

Add a cli to sidecar called `load-contracts` that will take a JSON blob and load contracts into the database to be used for indexing

### LINKED ISSUE #223: feat: dynamically upgrade side-loaded contract implementations

serichoi65:
## Description

This PR adds a new cli `load-contract` to allow users to add any contracts to Sidecar. When contracts are added by this command, they will be considered as non-core contracts, a.k.a `external` contracts. External contracts will automatically detect upgradeable contracts.

This is the list of flags associated to `load-contract`:
- `contract.address` (required)
    - `contract.abi` (required with address)
    - `contract.bytecode-hash` (optional)

- `contract.associate-to-proxy` (optional)
    - `contract.block-number` (required with associate-to-proxy)

- `contract.batch` (required to set to true when batch-loading, using file or stdin)
- `contract.from-file` (json file path)

Please refer to some examples below.

## Type of change

### Load a contract
```
% go build -o sidecar
% ./sidecar load-contract contract.address=0x870679e138bcdf293b7ff14dd44b70fc97e12fc0 --contract.abi='[{"input": []}]' --ethereum.rpc-url="http://72.46.85.253:8545" --chain="mainnet" --database.host=localhost --database.port="5432" --database.user=<add user> --database.password=<add password>  --database.db_name="sidecar"
```
Result:
```
sidecar=# select * from contracts where contract_address='0x870679e138bcdf293b7ff14dd44b70fc97e12fc0';
 id |              contract_address              |  contract_abi   |                          bytecode_hash                           | verified | matching_contract_address | checked_for_proxy | checked_for_abi |        created_at         |        updated_at         |     deleted_at      | contract_type 
----+--------------------------------------------+-----------------+------------------------------------------------------------------+----------+---------------------------+-------------------+-----------------+---------------------------+---------------------------+---------------------+---------------
 28 | 0x870679e138bcdf293b7ff14dd44b70fc97e12fc0 | [{"input": []}] | 6af45b3334e299fa97ea4781109e201a1411aba5131a6f1bc60fdf2011fce868 | t        |                           | f                 | t               | 2025-03-05 15:43:16.67865 | 2025-03-05 15:43:16.67865 | 0001-01-01 00:00:00 | external
(1 row)
```
### Load a proxy contract
```
% go build -o sidecar
% ./sidecar load-contract --contract.address=0xf5fd25a90902c27068cf5ebe53be8da693ac899e --contract.abi='[{"output": []}]' -- --contract.associate-to-proxy=0x870679e138bcdf293b7ff14dd44b70fc97e12fc0 --contract.block-number=19592323 --ethereum.rpc-url="http://72.46.85.253:8545" --chain="mainnet" --database.host=localhost --database.port="5432" --database.user=<add user> --database.password=<add password>  --database.db_name="sidecar"
```

Confirmed the query results:
```
sidecar=# select * from proxy_contracts where contract_address='0x870679e138bcdf293b7ff14dd44b70fc97e12fc0';
 block_number |              contract_address              |           proxy_contract_address           |         created_at         |         updated_at         |     deleted_at      
--------------+--------------------------------------------+--------------------------------------------+----------------------------+----------------------------+---------------------
     19592323 | 0x870679e138bcdf293b7ff14dd44b70fc97e12fc0 | 0xf5fd25a90902c27068cf5ebe53be8da693ac899e | 2025-03-05 15:43:16.744526 | 2025-03-05 15:43:16.744526 | 0001-01-01 00:00:00
(1 row)


sidecar=# select * from contracts where contract_address='0xf5fd25a90902c27068cf5ebe53be8da693ac899e';
 id |              contract_address              |   contract_abi   |                          bytecode_hash                           | verified | matching_contract_address | checked_for_proxy | checked_for_abi |         created_at         |         updated_at         |     deleted_at      | contract_type 
----+--------------------------------------------+------------------+------------------------------------------------------------------+----------+---------------------------+-------------------+-----------------+----------------------------+----------------------------+---------------------+---------------
 29 | 0xf5fd25a90902c27068cf5ebe53be8da693ac899e | [{"output": []}] | f249d9703e9baeeaa2e6b3167d5e71f4cc82127d02b24a955e978f8e3a5beb1b | t        |                           | f                 | t               | 2025-03-05 15:43:16.743142 | 2025-03-05 15:43:16.743142 | 0001-01-01 00:00:00 | external
(1 row)
```

### To batch load contracts
using `test.json`
```
{
	"core_contracts": [
		{
			"contract_address": "0x00a5fd09f6cee6ae9c8b0e5e33287f7c82880505",
			"contract_abi": "[{\"tests\": []}]",
			"bytecode_hash": "abc"
		},
		{
			"contract_address": "0x5d0b9ce2e277daf508528e9f6bf6314e79e4ed2b",
			"contract_abi": "[{\"tests2\": []}]",
			"bytecode_hash": "def"
		}
	],
	"proxy_contracts": [
		{
			"contract_address": "0x00a5fd09f6cee6ae9c8b0e5e33287f7c82880505",
			"proxy_contract_address": "0x5d0b9ce2e277daf508528e9f6bf6314e79e4ed2b",
			"block_number": 19592323
		}
	]
}
```

Command:
```
./sidecar load-contracts --ethereum.rpc-url="http://72.46.85.253:8545" --chain="mainnet" --database.host=localhost --database.port="5432" --database.user=<add user> --database.password=<add password> --database.db_name="sidecar" --contract.from-file=./test.json
```
**OR**
```
cat ./test.json | ./sidecar load-contracts --ethereum.rpc-url="http://72.46.85.253:8545" --chain="mainnet" --database.host=localhost --database.port="5432" --database.user=<add user> --database.password=<add password> --database.db_name="sidecar"
```

## Checklist:

### DIFF
--- a/cmd/loadContract.go
+++ b/cmd/loadContract.go
@@ -1,0 +1,184 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/Layr-Labs/sidecar/pkg/abiFetcher"
+	"github.com/Layr-Labs/sidecar/pkg/abiSource"
+	"github.com/Layr-Labs/sidecar/pkg/clients/ethereum"
+	"github.com/Layr-Labs/sidecar/pkg/contractStore"
+	"github.com/Layr-Labs/sidecar/pkg/contractStore/postgresContractStore"
+	"github.com/Layr-Labs/sidecar/pkg/postgres"
+	"github.com/Layr-Labs/sidecar/pkg/postgres/helpers"
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	"github.com/Layr-Labs/sidecar/pkg/postgres/migrations"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"gorm.io/gorm"
+)
+
+var loadContractCmd = &cobra.Command{
+	Use:   "load-contract [file]",
+	Short: "Load a contract",
+	Long:  `Load a contract from a file or stdin. If a file path is provided as the last argument, it will be used. If no file is provided, it will attempt to read from stdin.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		initLoadContractCmd(cmd)
+		cfg := config.NewConfig()
+
+		ctx := context.Background()
+
+		l, err := logger.NewLogger(&logger.LoggerConfig{Debug: cfg.Debug})
+		if err != nil {
+			return fmt.Errorf("failed to initialize logger: %w", err)
+		}
+
+		client := ethereum.NewClient(ethereum.ConvertGlobalConfigToEthereumConfig(&cfg.EthereumRpcConfig), l)
+
+		af := abiFetcher.NewAbiFetcher(client, abiFetcher.DefaultHttpClient(), l, cfg, []abiSource.AbiSource{})
+
+		pgConfig := postgres.PostgresConfigFromDbConfig(&cfg.DatabaseConfig)
+
+		pg, err := postgres.NewPostgres(pgConfig)
+		if err != nil {
+			return fmt.Errorf("failed to setup postgres connection: %w", err)
+		}
+
+		grm, err := postgres.NewGormFromPostgresConnection(pg.Db)
+		if err != nil {
+			return fmt.Errorf("failed to create gorm instance: %w", err)
+		}
+
+		migrator := migrations.NewMigrator(pg.Db, grm, l, cfg)
+		if err = migrator.MigrateAll(); err != nil {
+			return fmt.Errorf("failed to migrate: %w", err)
+		}
+
+		cs := postgresContractStore.NewPostgresContractStore(grm, l, cfg)
+
+		var filename string
+		var useFile bool
+		var useStdin bool
+
+		// Check if a file path is provided as a positional argument
+		if len(args) > 0 {
+			filename = args[0]
+			useFile = true
+		} else if cfg.LoadContractConfig.FromFile != "" {
+			// Check if a file path is provided via the --from-file flag
+			filename = cfg.LoadContractConfig.FromFile
+			useFile = true
+		} else {
+			// Check if we should read from stdin (if stdin is not a terminal)
+			stdinInfo, err := os.Stdin.Stat()
+			if err == nil && (stdinInfo.Mode()&os.ModeCharDevice) == 0 {
+				useStdin = true
+			}
+		}
+
+		// Process based on input source
+		if cfg.LoadContractConfig.Batch {
+			if useFile {
+				// Load contracts from file
+				err := cs.InitializeExternalContracts(filename)
+				if err != nil {
+					return fmt.Errorf("failed to initialize external contracts from file: %w", err)
+				}
+				return nil
+			}
+			if useStdin {
+				// Read directly from stdin
+				err := cs.InitializeExternalContractsFromReader(os.Stdin)
+				if err != nil {
+					return fmt.Errorf("failed to initialize external contracts from stdin: %w", err)
+				}
+				return nil
+			}
+			// If batch mode is enabled but no input source is provided
+			return fmt.Errorf("batch mode requires a file path or stdin input")
+		} else {
+			// If no file or stdin is provided, use the individual contract parameters
+			blockNumber := cfg.LoadContractConfig.BlockNumber
+			address := cfg.LoadContractConfig.Address
+			abi := cfg.LoadContractConfig.Abi
+			bytecodeHash := cfg.LoadContractConfig.BytecodeHash
+			associateToProxy := cfg.LoadContractConfig.AssociateToProxy
+
+			// Validate required parameters
+			if address == "" {
+				return fmt.Errorf("contract address is required in --contract.address")
+			}
+			if abi == "" {
+				return fmt.Errorf("contract ABI is required with the contract address in --contract.abi")
+			}
+			// Fetch bytecode hash if not provided
+			if bytecodeHash == "" {
+				bytecodeHash, err = af.FetchContractBytecodeHash(ctx, address)
+				if err != nil {
+					return fmt.Errorf("failed to fetch contract bytecode hash: %w", err)
+				}
+			}
+
+			// Wrap contract creation and proxy association in a transaction
+			_, err = helpers.WrapTxAndCommit[*contractStore.Contract](func(tx *gorm.DB) (*contractStore.Contract, error) {
+				// Create a temporary contract store with the transaction
+				txContractStore := postgresContractStore.NewPostgresContractStore(tx, l, cfg)
+
+				// Load the contract
+				contract, err := txContractStore.CreateContract(
+					address,
+					abi,
+					true,
+					bytecodeHash,
+					"",
+					true,
+					contractStore.ContractType_External,
+				)
+				if err != nil {
+					return nil, fmt.Errorf("failed to load contract: %w", err)
+				}
+
+				// Associate to proxy if specified
+				if associateToProxy != "" {
+					proxyContract, err := txContractStore.GetContractForAddress(associateToProxy)
+					if err != nil {
+						return nil, fmt.Errorf("error checking for proxy contract: %w", err)
+					}
+					if proxyContract == nil {
+						return nil, fmt.Errorf("proxy contract %s not found", associateToProxy)
+					}
+					if blockNumber == 0 {
+						return nil, fmt.Errorf("block number is required to load a proxy contract")
+					}
+					_, err = txContractStore.CreateProxyContract(blockNumber, associateToProxy, address)
+					if err != nil {
+						return nil, fmt.Errorf("failed to load to proxy contract: %w", err)
+					}
+				}
+
+				return contract, nil
+			}, grm, nil)
+
+			if err != nil {
+				return fmt.Errorf("failed to load contract with transaction: %w", err)
+			}
+
+			return nil
+		}
+	},
+}
+
+func initLoadContractCmd(cmd *cobra.Command) {
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		if err := viper.BindPFlag(config.KebabToSnakeCase(f.Name), f); err != nil {
+			fmt.Printf("Failed to bind flag '%s' - %+v\n", f.Name, err)
+		}
+		if err := viper.BindEnv(f.Name); err != nil {
+			fmt.Printf("Failed to bind env '%s' - %+v\n", f.Name, err)
+		}
+	})
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -67,6 +67,7 @@ func init() {
 	rootCmd.AddCommand(createSnapshotCmd)
 	rootCmd.AddCommand(restoreSnapshotCmd)
 	rootCmd.AddCommand(rpcCmd)
+	rootCmd.AddCommand(loadContractCmd)
 
 	// bind any subcommand flags
 	createSnapshotCmd.PersistentFlags().String(config.SnapshotOutputFile, "", "(deprecated, use --output) Path to save the snapshot file")
@@ -82,6 +83,14 @@ func init() {
 	restoreSnapshotCmd.PersistentFlags().String(config.SnapshotKind, "full", "The kind of snapshot to restore (slim, full, or archive)")
 
 	rpcCmd.PersistentFlags().String(config.SidecarPrimaryUrl, "", `RPC url of the "primary" Sidecar instance in an HA environment`)
+
+	loadContractCmd.PersistentFlags().String(config.LoadContractAddress, "", "Contract address to load")
+	loadContractCmd.PersistentFlags().String(config.LoadContractAbi, "", "ABI for the contract")
+	loadContractCmd.PersistentFlags().String(config.LoadContractBytecodeHash, "", "Bytecode hash for the contract")
+	loadContractCmd.PersistentFlags().String(config.LoadContractAssociateToProxy, "", "Contract address to associate to the proxy")
+	loadContractCmd.PersistentFlags().Uint64(config.LoadContractBlockNumber, uint64(0), "Block number for proxy contract deployment")
+	loadContractCmd.PersistentFlags().Bool(config.LoadContractBatch, false, "Batch load contracts from a JSON file")
+	loadContractCmd.PersistentFlags().String(config.LoadContractFromFile, "", "Path to a JSON file containing contract addresses to load")
 
 	rootCmd.PersistentFlags().VisitAll(func(f *pflag.Flag) {
 		key := config.KebabToSnakeCase(f.Name)
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -3,10 +3,14 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"log"
+	"time"
+
 	"github.com/Layr-Labs/sidecar/internal/metrics/prometheus"
 	"github.com/Layr-Labs/sidecar/internal/version"
 	"github.com/Layr-Labs/sidecar/pkg/abiFetcher"
 	"github.com/Layr-Labs/sidecar/pkg/abiSource"
+	"github.com/Layr-Labs/sidecar/pkg/abiSource/ipfs"
 	"github.com/Layr-Labs/sidecar/pkg/clients/ethereum"
 	sidecarClient "github.com/Layr-Labs/sidecar/pkg/clients/sidecar"
 	"github.com/Layr-Labs/sidecar/pkg/contractCaller/sequentialContractCaller"
@@ -30,9 +34,6 @@ import (
 	"github.com/Layr-Labs/sidecar/pkg/shutdown"
 	"github.com/Layr-Labs/sidecar/pkg/sidecar"
 	pgStorage "github.com/Layr-Labs/sidecar/pkg/storage/postgres"
-	"log"
-	"net/http"
-	"time"
 
 	"github.com/Layr-Labs/sidecar/internal/config"
 	"github.com/Layr-Labs/sidecar/internal/logger"
@@ -77,7 +78,8 @@ var runCmd = &cobra.Command{
 
 		client := ethereum.NewClient(ethereum.ConvertGlobalConfigToEthereumConfig(&cfg.EthereumRpcConfig), l)
 
-		af := abiFetcher.NewAbiFetcher(client, &http.Client{Timeout: 5 * time.Second}, l, cfg, []abiSource.AbiSource{})
+		ipfs := ipfs.NewIpfs(ipfs.DefaultHttpClient(), l, cfg)
+		af := abiFetcher.NewAbiFetcher(client, abiFetcher.DefaultHttpClient(), l, cfg, []abiSource.AbiSource{ipfs})
 
 		pgConfig := postgres.PostgresConfigFromDbConfig(&cfg.DatabaseConfig)
 
@@ -140,7 +142,7 @@ var runCmd = &cobra.Command{
 
 		go rcq.Process()
 
-		p := pipeline.NewPipeline(fetchr, idxr, mds, sm, msm, rc, rcq, cfg, sink, eb, l)
+		p := pipeline.NewPipeline(fetchr, idxr, mds, contractStore, cm, sm, msm, rc, rcq, cfg, sink, eb, l)
 
 		scc, err := sidecarClient.NewSidecarClient(cfg.SidecarPrimaryConfig.Url, !cfg.SidecarPrimaryConfig.Secure)
 		if err != nil {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -117,6 +117,16 @@ type SidecarPrimaryConfig struct {
 	IsPrimary bool
 }
 
+type LoadContractConfig struct {
+	Address          string
+	Abi              string
+	BytecodeHash     string
+	AssociateToProxy string
+	BlockNumber      uint64
+	Batch            bool
+	FromFile         string
+}
+
 type IpfsConfig struct {
 	Url string
 }
@@ -137,6 +147,7 @@ type Config struct {
 	DataDogConfig         DataDogConfig
 	PrometheusConfig      PrometheusConfig
 	SidecarPrimaryConfig  SidecarPrimaryConfig
+	LoadContractConfig    LoadContractConfig
 	IpfsConfig            IpfsConfig
 	EtherscanConfig       EtherscanConfig
 }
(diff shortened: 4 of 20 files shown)


cmd/sideloadContract.go
```diff
@@ -0,0 +1,196 @@
...
+			return fmt.Errorf("failed to create gorm instance: %w", err)
+		}
+
+		migrator := migrations.NewMigrator(pg.Db, grm, l, cfg)
+		if err = migrator.MigrateAll(); err != nil {
+			return fmt.Errorf("failed to migrate: %w", err)
+		}
+
+		contractStore := postgresContractStore.NewPostgresContractStore(grm, l, cfg)
+		// if err := contractStore.InitializeCoreContracts(); err != nil {
```
serichoi65, line 94: Skipping initialize core contracts during `Sideload`. Will remove if this logic is confirmed.

cmd/sideloadContract.go
```diff
@@ -0,0 +1,196 @@
...
+			pServer := prometheus.NewPrometheusServer(&prometheus.PrometheusServerConfig{
+				Port: cfg.PrometheusConfig.Port,
+			}, l)
+			if err := pServer.Start(promChan); err != nil {
+				return fmt.Errorf("failed to start prometheus server: %w", err)
+			}
+		}
+
+		// Start the sidecar main process in a goroutine so that we can listen for a shutdown signal
+		go sidecar.StartSideload(ctx)
```
serichoi65, line 171: This file is similar to `run.go` but is only differentiated here.

cmd/root.go
```diff
@@ -82,6 +83,14 @@ func init() {
 
 	rpcCmd.PersistentFlags().String(config.SidecarPrimaryUrl, "", `RPC url of the "primary" Sidecar instance in an HA environment`)
 
+	sideloadContractCmd.PersistentFlags().Bool(config.SideloadEnabled, false, "Enable contract side-loading")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadAbiSource, "", "ABI source for the contract (etherscan, ipfs)")
```
seanmcgary: For loading a contract directly, I dont think we need it. They should be providing the ABI directly and only _new_ ABIs from upgrades are pulled from either etherscan or IPFS

cmd/root.go
```diff
@@ -82,6 +83,14 @@ func init() {
 
 	rpcCmd.PersistentFlags().String(config.SidecarPrimaryUrl, "", `RPC url of the "primary" Sidecar instance in an HA environment`)
 
+	sideloadContractCmd.PersistentFlags().Bool(config.SideloadEnabled, false, "Enable contract side-loading")
```
seanmcgary: I mentioned this on Slack, but rather than having a flag to enable contract sideloading, we should add a column to the contracts table called like `contract_type` which is an `enum(core, external)`. This way the enablement of side-loaded contracts is dictated by if you've loaded any or not

cmd/root.go
```diff
@@ -82,6 +83,14 @@ func init() {
...
+	sideloadContractCmd.PersistentFlags().String(config.SideloadAbiSource, "", "ABI source for the contract (etherscan, ipfs)")
+	sideloadContractCmd.PersistentFlags().String(config.IpfsUrl, "", "IPFS gateway URL")
+	sideloadContractCmd.PersistentFlags().String(config.EtherscanApiKey, "", "Etherscan API key")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadContractAddress, "", "Contract address to side-load")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadProxyContractAddress, "", "Proxy contract address for the contract")
+	sideloadContractCmd.PersistentFlags().Uint64(config.SideloadBlockNumber, uint64(0), "Block number for proxy contract deployment")
```
seanmcgary: Like the above comment, we should take the ABI directly here. Taking it via `stdin` is probably the most flexible option

cmd/root.go
```diff
@@ -82,6 +83,14 @@ func init() {
 
 	rpcCmd.PersistentFlags().String(config.SidecarPrimaryUrl, "", `RPC url of the "primary" Sidecar instance in an HA environment`)
 
+	sideloadContractCmd.PersistentFlags().Bool(config.SideloadEnabled, false, "Enable contract side-loading")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadAbiSource, "", "ABI source for the contract (etherscan, ipfs)")
+	sideloadContractCmd.PersistentFlags().String(config.IpfsUrl, "", "IPFS gateway URL")
+	sideloadContractCmd.PersistentFlags().String(config.EtherscanApiKey, "", "Etherscan API key")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadContractAddress, "", "Contract address to side-load")
+	sideloadContractCmd.PersistentFlags().String(config.SideloadProxyContractAddress, "", "Proxy contract address for the contract")
```
seanmcgary: Our own naming for `contractAddress` and `proxyContractAddress` isnt the most clear.

What if we went with `contractAddress` and then an optional `isImplementationForAddress` flag to denote that you're loading an implementation rather than a top-level proxy?

seanmcgary (changes requested):

cmd/batchLoadContracts.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	"github.com/Layr-Labs/sidecar/pkg/postgres/migrations"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+var batchLoadContractsCmd = &cobra.Command{
+	Use:   "batch-load-contracts",
```
seanmcgary: Rather than creating a separate command, could we just include a `--from-file` argument in `load-contract`?

cmd/batchLoadContracts.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		migrator := migrations.NewMigrator(pg.Db, grm, l, cfg)
+		if err = migrator.MigrateAll(); err != nil {
+			return fmt.Errorf("failed to migrate: %w", err)
+		}
+
+		contractStore := postgresContractStore.NewPostgresContractStore(grm, l, cfg)
+
+		filename := cfg.BatchLoadContractsConfig.JsonFile
```
seanmcgary: Rather than specifying a flag, this could assume the last argument when invoking `batch-load-contracts` is the file to use (if present) OR it should attempt to read from `stdin`. That way your UX looks like:

```
sidecar batch-load-contracts --flags... ~/my-contracts.json

cat ~/my-contracts.json | sidecar batch-load-contracts --flags...
```

cmd/loadContract.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/Layr-Labs/sidecar/internal/logger"
+	"github.com/Layr-Labs/sidecar/internal/metrics"
+	"github.com/Layr-Labs/sidecar/pkg/postgres/migrations"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+var loadContractCmd = &cobra.Command{
+	Use:   "load-contract",
```
seanmcgary: Yeah, we should definitely roll the functionality of `batch-load-contracts` into here. Taking a `--batch` boolean flag to ignore the singular contract flags then taking the file as the last argument or stdin would work

cmd/loadContract.go
```diff
@@ -0,0 +1,99 @@
...
+
+		cm := contractManager.NewContractManager(contractStore, client, af, sink, l)
+
+		blockNumber := cfg.LoadContractConfig.BlockNumber
+		contractAddress := cfg.LoadContractConfig.ContractAddress
+		contractAbi := cfg.LoadContractConfig.ContractAbi
+		implementationForAddress := cfg.LoadContractConfig.ImplementationForAddress
+		implementationAbi := cfg.LoadContractConfig.ImplementationAbi
+
+		err = cm.InitializeLoadingContract(ctx, blockNumber, contractAddress, contractAbi, implementationForAddress, implementationAbi)
```
seanmcgary: Havent figured out a linting rule for this yet, but long lines/function calls with > 3 arguments should be wrapped to new lines, which also means you can get rid of the intermediate assignments above:

```go
err = cm.InitializeLoadingContract(
    ctx,
    cfg.LoadContractConfig.BlockNumber,
    cfg.LoadContractConfig.ContractAddress,
    cfg.LoadContractConfig.ContractAbi,
    cfg.LoadContractConfig.ImplementationForAddress,
    cfg.LoadContractConfig.ImplementationAbi,
)
```

The other option would be to take the contract information as a struct

cmd/run.go
```diff
@@ -77,7 +78,8 @@ var runCmd = &cobra.Command{
 
 		client := ethereum.NewClient(ethereum.ConvertGlobalConfigToEthereumConfig(&cfg.EthereumRpcConfig), l)
 
-		af := abiFetcher.NewAbiFetcher(client, &http.Client{Timeout: 5 * time.Second}, l, cfg, []abiSource.AbiSource{})
+		ipfs := ipfs.NewIpfs(&http.Client{Timeout: 5 * time.Second}, l, cfg)
```
seanmcgary: QOL suggestion: probably worth adding a `DefaultHttpClient` function to the ipfs and abiFetcher modules that returns `&http.Client{Timeout: 5 * time.Second}`.

internal/config/config.go
```diff
@@ -199,6 +213,14 @@ var (
 
 	SidecarPrimaryUrl = "sidecar-primary.url"
 
+	LoadContractAddress                  = "load-contract.contract-address"
```
seanmcgary: Since these flags only exist on the `load-contract` command, theres no need to prefix them with `load-contract.`.

I would probably organize them as:

```
--contract.address
--contract.abi
--contract.implementation-for-address
--contract.implementaiton-abi
--contract.block-number
```

pkg/contractManager/contractManager.go
```diff
@@ -120,10 +120,6 @@ func (cm *ContractManager) HandleContractUpgrade(ctx context.Context, blockNumbe
 	return nil
 }
 
-// CreateUpgradedProxyContract creates a new proxy contract relationship in the database.
-// It creates entries for both the proxy contract and the implementation contract,
-// fetching the ABI for the implementation contract.
-// If the proxy contract already exists, it returns without error.
```
seanmcgary: doc comment should be added back or updated to reflect anything new/modified

pkg/contractManager/contractManager.go
```diff
@@ -171,13 +177,88 @@ func (cm *ContractManager) CreateUpgradedProxyContract(
...
+	if err != nil {
+		cm.Logger.Sugar().Errorw("Failed to create a contract for contractAddress",
+			zap.Error(err),
+			zap.String("address", contractAddress),
+		)
+		return err
+	}
+
+	// create a contract for implementationForAddress
+	implementationBytecodeHash, err := cm.AbiFetcher.FetchContractBytecodeHash(ctx, implementationForAddress)
```
seanmcgary: We're passing `implementationAbi` as an argument, so do we need to call this at all?

pkg/contractManager/contractManager.go
```diff
@@ -171,13 +177,88 @@ func (cm *ContractManager) CreateUpgradedProxyContract(
...
+func (cm *ContractManager) InitializeLoadingContract(
+	ctx context.Context,
+	blockNumber uint64,
+	contractAddress string,
+	contractAbi string,
+	implementationForAddress string,
+	implementationAbi string,
+) error {
+	// create a contract for contractAddress
+	contractBytecodeHash, err := cm.AbiFetcher.FetchContractBytecodeHash(ctx, contractAddress)
```
seanmcgary: We should probably take the bytecode hash as an argument when adding a contract, or at least make it optional and then fetch it if not present.

Zeus outputs the bytecode hash that we've been using in the static files, so at least for our contracts we can provide it directly.

cmd/loadContract.go
seanmcgary: Also what if any or all of these flags are empty? We need some error checking on them.

I would also suggest decoupling creating the contract and proxy association. Right now its a little ambiguous as to how Im supposed to sequence loading something, especially when my proxy has multiple implementations. For example, if Im trying to load the rewards coordinator, it should probably go something like:

* Load the proxy contract (implementation flags are empty)
* For each implementation, load the implementation.
  * `--contract-address` and `--contract-abi` would be the address and abi of the implementation
  * `--implementation-for-address` would be the address of the top-level proxy
  * `--implementation-abi` would (and should) be empty since I already loaded it

pkg/contractManager/contractManager.go
```diff
@@ -171,13 +177,88 @@ func (cm *ContractManager) CreateUpgradedProxyContract(
...
 		)
 		return err
 	}
+
 	cm.Logger.Sugar().Debugf("Created new contract for proxy contract", zap.String("proxyContractAddress", proxyContractAddress))
 
 	return nil
 }
+
+func (cm *ContractManager) InitializeLoadingContract(
```
seanmcgary: `LoadContract` is a better name

pkg/contractStore/postgresContractStore/postgresContractStore.go
```diff
@@ -83,10 +85,18 @@ func (s *PostgresContractStore) CreateContract(
 	bytecodeHash string,
 	matchingContractAddress string,
 	checkedForAbi bool,
+	contractType ...string,
```
seanmcgary: This shouldnt be a spread string; It should be explicitly required and set to a constant.  Making it a spread implies I can pass multiple values which isnt the case.

I would create two constants:

```go
type ContractType string

const (
    ContractType_Core ContractType = "core"
    ContractType_External ContractType = "external"
)
```

and then use them in each appropriate place where `CreateContract` is currently called.

pkg/contractStore/postgresContractStore/postgresContractStore.go
```diff
@@ -110,7 +120,14 @@ func (s *PostgresContractStore) FindOrCreateContract(
...
```
seanmcgary: Same here, see above comment

pkg/contractStore/postgresContractStore/postgresContractStore.go
```diff
@@ -349,3 +363,35 @@ func (s *PostgresContractStore) InitializeCoreContracts() error {
...
+	}
+
+	if err := s.InitializeContracts(coreContracts, "core"); err != nil {
+		return fmt.Errorf("failed to initialize core contracts: %w", err)
+	}
+	return nil
+}
+
+func (s *PostgresContractStore) InitializeExternalContracts(filename string) error {
+	jsonData, err := os.ReadFile(filename)
```
seanmcgary: This should be decoupled from actually reading the file. In fact, you could refactor this and the core contracts one above to be a more generic `LoadContractsFromJson(data string, contractType string)`.

This would support reading from a file on disk and taking the contents from stdin for batch loading.

pkg/indexer/indexer.go
```diff
@@ -235,7 +235,12 @@ func (idx *Indexer) IsInterestingAddress(addr string) bool {
 	if addr == "" {
 		return false
 	}
-	return slices.Contains(idx.Config.GetInterestingAddressForConfigEnv(), strings.ToLower(addr))
+	isExternalInteresting := false
+	contract, _ := idx.ContractStore.GetContractForAddress(addr)
```
seanmcgary: Since we use the top-level proxy contract address to determine if something is interesting, I would change this whole function to read from the database.

```
addresses, err := idx.ContractStore.GetProxyAddresses()
if err != nil {
    return false, err
}
return slices.Contains(addresses, strings.ToLower(addr)), nil
```

`GetProxyAddresses` would do a select distinct on the proxy_contracts table to find all top-level proxies.

seanmcgary (changes requested):

pkg/contractStore/postgresContractStore/postgresContractStore.go
serichoi65: Created a more generic `InitializeExternalContractsFromReader(reader io.Reader)`

cmd/loadContract.go
serichoi65: Kept the intermediate assignments now that there's a check added for flag logics.

cmd/batchLoadContracts.go
serichoi65: Both commands now work! (but now with `load-contracts`, as `batch-load-contracts` is deprecated)

cmd/loadContract.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			// Check if we should read from stdin (if stdin is not a terminal)
+			stdinInfo, err := os.Stdin.Stat()
+			if err == nil && (stdinInfo.Mode()&os.ModeCharDevice) == 0 {
+				useStdin = true
+			}
+		}
+
+		// Process based on input source
+		if cfg.LoadContractConfig.Batch && useFile {
```
seanmcgary: Suggestion to make this a bit neater and cut down on the `else if`'s:

```go
// Process based on input source
if cfg.LoadContractConfig.Batch {
	if useFile {
		// Load contracts from file
		err := contractStore.InitializeExternalContracts(filename)
		if err != nil {
			return fmt.Errorf("failed to initialize external contracts from file: %w", err)
		}
		return nil
	}
	if useStdin {
		// Read directly from stdin
		err := contractStore.InitializeExternalContractsFromReader(os.Stdin)
		if err != nil {
			return fmt.Errorf("failed to initialize external contracts from stdin: %w", err)
		}
		return nil
	}
	return fmt.Errorf("batch mode requires a file path or stdin input")
}

// everything else from the last "else"
```

cmd/loadContract.go
```diff
@@ -0,0 +1,166 @@
...
+			abi := cfg.LoadContractConfig.Abi
+			bytecodeHash := cfg.LoadContractConfig.BytecodeHash
+			associateToProxy := cfg.LoadContractConfig.AssociateToProxy
+
+			// Validate required parameters
+			if address == "" {
+				return fmt.Errorf("contract address is required in --contract.address")
+			} else if abi == "" {
+				return fmt.Errorf("contract ABI is required with the contract address in --contract.abi")
+			}
```
seanmcgary, line 117: Stylistically I would make these two distinct statements to make it easier to read. 

```go
if address == "" {
    return fmt.Errorf("contract address is required in --contract.address")
}
if abi == "" {
	return fmt.Errorf("contract ABI is required with the contract address in --contract.abi")
}
```

cmd/loadContract.go
```diff
@@ -0,0 +1,166 @@
...
+				bytecodeHash, err = af.FetchContractBytecodeHash(ctx, address)
+				if err != nil {
+					return fmt.Errorf("failed to fetch contract bytecode hash: %w", err)
+				}
+			}
+			// Associate to proxy if specified
+			if associateToProxy != "" {
+				_, err := contractStore.GetContractForAddress(associateToProxy)
+				if err != nil {
+					return fmt.Errorf("proxy contract %s not found: %w", associateToProxy, err)
```
seanmcgary: If you look carefully at `GetContractForAddress`, we actually return `nil, nil` for a `ErrRecordNotFound` response, so the error in this case is something else. You'll need to add a check to see that the returned contract is `nil` to satisfy the not found vs. found check

seanmcgary: If it is nil, I would return an error telling the caller that the proxy they want to associate to doesnt exist

cmd/loadContract.go
```diff
@@ -0,0 +1,166 @@
...
+				}
+				if blockNumber == 0 {
+					return fmt.Errorf("block number is required to load a proxy contract")
+				}
+				err = cm.LoadProxyContract(ctx, blockNumber, associateToProxy, address)
```
seanmcgary: Technically this should come after calling `cm.LoadContract`. I totally thought we had a fk constraint on the `proxy_contract` table, but I guess we dont, otherwise this would have failed.

I would also look into wrapping this in a transaction so that if one fails, the whole operation rolls back.

See this for an example of using `WrapTxAndCommit`

https://github.com/Layr-Labs/sidecar/blob/master/pkg/contractStore/postgresContractStore/postgresContractStore.go#L115-L136

pkg/contractManager/contractManager.go
```diff
@@ -169,15 +179,64 @@ func (cm *ContractManager) CreateUpgradedProxyContract(
...
 		)
 		return err
 	}
+
 	cm.Logger.Sugar().Debugf("Created new contract for proxy contract", zap.String("proxyContractAddress", proxyContractAddress))
 
 	return nil
 }
+
+func (cm *ContractManager) LoadContract(
```
seanmcgary: Since this just turns around and calls `cm.ContractStore.CreateContract`, can we skip this function all together and just call the contract store where we call `LoadContract`?

pkg/contractManager/contractManager.go
```diff
@@ -169,15 +179,64 @@ func (cm *ContractManager) CreateUpgradedProxyContract(
...
+			zap.Error(err),
+			zap.String("address", address),
+		)
+		return err
+	}
+
+	return nil
+}
+
+func (cm *ContractManager) LoadProxyContract(
```
seanmcgary: Same as with `LoadContract` can we just call `cm.ContractStore.CreateProxyContract` directly?

pkg/contractStore/postgresContractStore/postgresContractStore.go
```diff
@@ -73,6 +76,33 @@ func (s *PostgresContractStore) GetProxyContractForAddress(blockNumber uint64, a
 	return proxyContract, nil
 }
 
+// GetAllProxyAddressesInString returns all proxy addresses in the database as a slice of strings.
+func (s *PostgresContractStore) GetAllProxyAddressesInString() ([]string, error) {
+	var addressPointers []*string
+
+	result := s.Db.Model(&contractStore.ProxyContract{}).
+		Distinct().
+		Pluck("contract_address", &addressPointers)
```
seanmcgary: Bit more ergonomic to do:

```
var addresses []string
res := s.Db.Raw(`select distinct(contract_address) from proxy_contracts`).Scan(&addresses)
if res.Error != nil {
    return res.Error
}
return addresses, nil
```

pkg/pipeline/pipeline.go
```diff
@@ -231,6 +240,24 @@ func (p *Pipeline) RunForFetchedBlock(ctx context.Context, block *fetcher.Fetche
 				hasError = true
 				return err
 			}
+
+			contract, err := p.contractStore.GetContractForAddress(strings.ToLower(log.Address))
```
seanmcgary, line 243: to prevent us from hitting the DB on every log, I would check to see if `log.EventName == "Upgraded"` first before querying the db

seanmcgary (changes requested):

cmd/loadContract.go
serichoi65: That makes sense that creating a contract should come first. I tested locally to make sure the rollback works properly.

cmd/loadContract.go
serichoi65: Only checking if the contract is nil since the more error message will also show from `GetContractForAddress`

seanmcgary (approved):

PR #298, opened 2025-03-11
feat: transaction log backfiller & log parse refactor

seanmcgary:
## Description

* Refactored transaction log parsing into a reusable module
* Added a new `TransactionBackfiller` construct

### TransactionBackfiller

The backfiller is a module that will handle fetching blocks, transactions and transactionReceipts. Each log is processed and handed to two functions:

* `IsInterestingLog`
* `TransactionLogHandler`

Both of these allow for custom implementations based on the implementers criteria rather rather than relying on the default behavior found in the `pipeline`. The purpose of the backfiller is two fold:

* Provide a means to backfill transaction logs when recognizing a new core contract
* Provide the same means as a library for anyone side-loading contracts that may need to re-index events from past blocks.

For now, this is implemented as a library and can be enabled in the Sidecar core when it becomes necessary to backfill past contract events due to a new core contract getting added.

## Type of change

## How Has This Been Tested?

Manual testing, added more automated tests

## Checklist:


cmd/debugger/main.go
```diff
@@ -156,21 +156,23 @@ func main() {
...
-	}
-
-	decodedLog, err := idxr.DecodeLogWithAbi(nil, receipt, interestingLog)
-	if err != nil {
-		l.Sugar().Fatalw("Failed to decode log", zap.Error(err))
-	}
-	l.Sugar().Infof("Decoded log: %+v", decodedLog)
+	fmt.Printf("Transaction: %+v\n", transaction)
+
+	//logIndex := 4
```
serichoi65: Uncomment or delete or keep the commented section?

pkg/transactionBackfiller/queue.go
```diff
@@ -0,0 +1 @@
+package transactionBackfiller
```
serichoi65: queue is already set up as `chan` in transactionBackfiller, right?

pkg/transactionLogParser/logParser.go
```diff
@@ -0,0 +1,309 @@
...
+			zap.Uint64("blockNumber", receipt.BlockNumber.Value()),
+		)
+		return nil, nil
+	}
+
+	var a *abi.ABI
+	logs := make([]*parser.DecodedLog, 0)
+
+	for i, lg := range receipt.Logs {
+		if !tlp.interestingLogQualifier.IsInterestingAddress(lg.Address.Value()) {
```
serichoi65, line 83: Do we override or create our own, like shown in backfiller_test? or I thought about taking the config from the backfill cli and add it to the existing `IsInterestingAddress`

cmd/debugger/main.go
seanmcgary: Debugger is kinda a dumping ground for random stuff which is why I left it there. I'll remove it though

pkg/transactionLogParser/logParser.go
seanmcgary: The function signature is the same as our other instances of `IsInterestingAddress`, so you could give it a reference to an existing function, give it a closure that then calls an existing function, or write a new one entirely.

pkg/transactionBackfiller/queue.go
seanmcgary: Removed this file since its not used

serichoi65 (approved):

PR #344, opened 2025-04-02
perf: implement a faster from-scratch sync

seanmcgary:
## Description

Use `eth_getLogs` to only get logs that are interesting to use. This allows us to skip fetching transaction receipts that we dont need saving us on RPC requests and data transfer.

## Type of change

- [x] Performance improvement

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #350, opened 2025-04-09
feat: option to backfill with specific addresses

serichoi65:
## Description

This adds a feature to `transactionBackfiller` to backfill specific addresses.

Fixes # (issue)

## Type of change

- [ ] Performance improvement

## How Has This Been Tested?

Added a unit test.

## Checklist:

### DIFF
--- a/pkg/transactionBackfiller/backfiller.go
+++ b/pkg/transactionBackfiller/backfiller.go
@@ -3,11 +3,12 @@ package transactionBackfiller
 import (
 	"context"
 	"fmt"
+	"sync"
+
 	"github.com/Layr-Labs/sidecar/pkg/clients/ethereum"
 	"github.com/Layr-Labs/sidecar/pkg/fetcher"
 	"github.com/Layr-Labs/sidecar/pkg/storage"
 	"go.uber.org/zap"
-	"sync"
 )
 
 // TransactionBackfillerConfig contains configuration options for the transaction backfiller.
@@ -28,6 +29,8 @@ type BackfillerMessage struct {
 	StartBlock uint64
 	// EndBlock is the last block to process (inclusive)
 	EndBlock uint64
+	// Addresses is a list of addresses to process
+	Addresses []string
 	// TransactionLogHandler is a function called for each interesting log
 	TransactionLogHandler func(block *ethereum.EthereumBlock, receipt *ethereum.EthereumTransactionReceipt, log *ethereum.EthereumEventLog) error
 	// IsInterestingLog determines if a log should be processed
@@ -47,6 +50,13 @@ type TransactionBackfiller struct {
 	blockStore storage.BlockStore
 	queue      chan *BackfillerMessage
 	done       chan struct{}
+}
+
+// LogWithMetadata represents a log with additional information needed for processing
+type LogWithMetadata struct {
+	Log         *ethereum.EthereumEventLog
+	BlockNumber uint64
+	TxHash      string
 }
 
 const (
@@ -196,6 +206,12 @@ func (t *TransactionBackfiller) Process() {
 //   - *BackfillerResponse: The result of processing the blocks
 func (t *TransactionBackfiller) ProcessBlocks(ctx context.Context, queueMessage *BackfillerMessage) *BackfillerResponse {
 	response := &BackfillerResponse{}
+
+	// If addresses are specified, use the optimized method
+	if len(queueMessage.Addresses) > 0 {
+		return t.ProcessBlocksWithAddresses(ctx, queueMessage)
+	}
+
 	wg := &sync.WaitGroup{}
 	blocksQueue := make(chan uint64, queueMessage.EndBlock-queueMessage.StartBlock+1)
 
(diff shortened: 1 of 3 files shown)


pkg/transactionBackfiller/backfiller.go
```diff
@@ -235,6 +251,222 @@ func (t *TransactionBackfiller) ProcessBlocks(ctx context.Context, queueMessage
...
+
+// fetchLogsForAddresses fetches logs for each address in parallel and sends them to the logsChan.
+// Returns any errors encountered during the fetch process.
+func (t *TransactionBackfiller) fetchLogsForAddresses(
+	ctx context.Context,
+	queueMessage *BackfillerMessage,
+	logsChan chan<- *LogWithMetadata,
+) []error {
+	var wgFetch sync.WaitGroup
+	var fetchMutex sync.Mutex
```
seanmcgary: Rather than using a mutex (which can kill your performance), you should aggregate results in a channel and then read from the channel to aggregate them together. This is the pattern we use pretty much everywhere we process things in parallel.

pkg/transactionBackfiller/backfiller.go
```diff
@@ -235,6 +251,222 @@ func (t *TransactionBackfiller) ProcessBlocks(ctx context.Context, queueMessage
...
+	if len(allErrors) > 0 {
+		response.Errors = allErrors
+	}
+
+	return response
+}
+
+// fetchLogsForAddresses fetches logs for each address in parallel and sends them to the logsChan.
+// Returns any errors encountered during the fetch process.
+func (t *TransactionBackfiller) fetchLogsForAddresses(
```
seanmcgary: I actually have this implemented over here [1] as part of the fetcher. You should be able to take it entirely if you want.

[1] https://github.com/Layr-Labs/sidecar/pull/344/files#diff-84c47d253e7b4669261ec0dd875884d7ff732bd0f4939b4697bfd0de55486ff8R238-R370

pkg/transactionBackfiller/backfiller.go
```diff
@@ -235,6 +251,222 @@ func (t *TransactionBackfiller) ProcessBlocks(ctx context.Context, queueMessage
...
+// processLogsBatch processes logs received via logsChan using a worker pool.
+// Returns any errors encountered during processing.
+func (t *TransactionBackfiller) processLogsBatch(
+	ctx context.Context,
+	queueMessage *BackfillerMessage,
+	logsChan <-chan *LogWithMetadata,
+	numWorkers int,
+) []error {
+	var wgProcess sync.WaitGroup
+	processMutex := sync.Mutex{}
```
seanmcgary: Same here - use the channel aggregation pattern rather than a mutex

pkg/transactionBackfiller/backfiller.go
```diff
@@ -235,6 +251,222 @@ func (t *TransactionBackfiller) ProcessBlocks(ctx context.Context, queueMessage
...
+	response := &BackfillerResponse{}
+
+	t.logger.Sugar().Infow("Processing blocks with specified addresses",
+		zap.Uint64("startBlock", queueMessage.StartBlock),
+		zap.Uint64("endBlock", queueMessage.EndBlock),
+		zap.Any("addresses", queueMessage.Addresses),
+	)
+
+	// A channel to collect all logs from all addresses
+	logsChan := make(chan *LogWithMetadata, 1000)
```
seanmcgary: I would look to simplify your fetching and parsing to not rely on passing around channels. The parallelization should happen at a very "top" level and be able to call a series of functions that dont require a channel to operator.

- First take and rename: `FetchLogsForContractsForBlockRange` to something like `FetchInterestingBlocksForContractsForBlockRange` (since thats technically more accurate...)
- Create new function `FetchLogsForContractsForRange` that contains the parallelization of `FetchInterestingBlocksForContractsForBlockRange` but returns the logs without filtering down to just the block numbers. `FetchInterestingBlocksForContractsForBlockRange` can now call this new function you just created.
- For each log in the results, process and index

Since the RPC calls are the expensive ones, thats really the only place you'll benefit from spawning multiple goroutines. Once you've fetched all of the interesting logs for the current range, processing them and inserting should be really fast. If you wanted to speed up insertion, I would aggregate all of the logs to insert together and execute a batch insert with Gorm

seanmcgary (changes requested):

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+//
+// Returns:
+//   - *BackfillerResponse: The result of processing the logs
+func (t *TransactionBackfiller) ProcessBlocksWithAddresses(ctx context.Context, queueMessage *BackfillerMessage) *BackfillerResponse {
+	response := &BackfillerResponse{}
+
+	interestingBlocks, logs, err := t.fetcher.FetchInterestingBlocksAndLogsForContractsForBlockRange(ctx, queueMessage.StartBlock, queueMessage.EndBlock, queueMessage.Addresses)
+	if err != nil {
+		t.logger.Sugar().Errorw("Error fetching interesting blocks and logs", zap.Error(err))
+		return response
```
seanmcgary: Rather than returning an empty struct, I would return the error

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+		blockNum := log.BlockNumber.Value()
+		txHash := log.TransactionHash.Value()
+
+		logWithMetadata := &LogWithMetadata{
+			Log:         log,
+			BlockNumber: blockNum,
+			TxHash:      txHash,
+		}
+
+		if _, exists := logsByBlock[blockNum]; !exists {
```
seanmcgary: The convention in Go is to use `ok` here rather than `exists`

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+		logsByBlock[blockNum] = append(logsByBlock[blockNum], logWithMetadata)
+	}
+
+	// Create worker pool for block fetching and processing
+	wg := &sync.WaitGroup{}
+	blockQueue := make(chan uint64, len(interestingBlocks))
+	errorsRecv := make(chan error, len(interestingBlocks))
+
+	// Launch workers
+	wg.Add(t.config.Workers)
```
seanmcgary, line 390: Good use of the worker queue

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+	for i := 0; i < t.config.Workers; i++ {
+		go func() {
+			defer wg.Done()
+			for blockNum := range blockQueue {
+				logsForBlock, exists := logsByBlock[blockNum]
```
seanmcgary: This is not threadsafe. Rather than queuing just the block numbers, queue the logs too. Probably need to create a struct to wrap the block number and logs to push into the queue if you need the block number below

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+	t.logger.Sugar().Infow("All workers finished, compiling errors")
+	close(errorsRecv)
+	errors := make([]error, 0)
+	for err := range errorsRecv {
+		errors = append(errors, err)
+	}
+
+	if len(errors) > 0 {
+		response.Errors = errors
+		return response
```
seanmcgary: no need for the pre-mature return since we're immediately returning `response` after the if statement

pkg/transactionBackfiller/backfiller.go
```diff
@@ -296,3 +312,127 @@ func (t *TransactionBackfiller) ProcessBlock(ctx context.Context, blockNumber ui
...
+
+	return response
+}
+
+// processLogsForBlock processes all logs for a specific block
+func (t *TransactionBackfiller) processLogsForBlock(
+	queueMessage *BackfillerMessage,
+	logs []*LogWithMetadata,
+	fetchedBlock *fetcher.FetchedBlock,
+	errorsChan chan<- error,
```
seanmcgary: Rather than dealing with this channel, this function can aggregate a list of errors and simply return it. Alternatively, you'd probably want to return the first error encountered, otherwise you may end up with partial data handled

seanmcgary (changes requested):

seanmcgary (approved):

PR #373, opened 2025-04-29
feat: calculate rewards daily given the flag, instead of using the cronjob

### LINKED ISSUE #315: feat: automatically generate rewards data daily
A common trend is that customers want rewards data calculated daily. Rather than require them to run a cronjob to probe the API, we should add a flag that does this automatically.

`--calculate-rewards-daily=true` would tell the block processor to check if a block is the first in a new day and if it is, queue a message to begin rewards calculation for that day. This would not run during the backfill process since it would cause backfilling to take too long.

serichoi65:
## Description

Currently, sidecar users have to run the cronjob to probe the API and calculate rewards daily on their own, but this PR provides a flag during indexing blocks to automatically does the job, given the `--rewards.calculate_rewards_daily` flag.

Fixes #315 

## Type of change

- [ ] Performance improvement

## How Has This Been Tested?

Tested the date logic of finding the first block past midnight locally.

## Checklist:

### DIFF
--- a/pkg/pipeline/pipeline.go
+++ b/pkg/pipeline/pipeline.go
@@ -121,6 +121,16 @@ func (p *Pipeline) RunForFetchedBlock(ctx context.Context, block *fetcher.Fetche
 		})
 	}()
 
+	// Check if we should calculate daily rewards
+	if p.globalConfig.Rewards.CalculateRewardsDaily && !isBackfill {
+		err := p.CalculateRewardsDaily(ctx, block)
+		if err != nil {
+			p.Logger.Sugar().Errorw("Failed to calculate daily rewards", zap.Uint64("blockNumber", blockNumber), zap.Error(err))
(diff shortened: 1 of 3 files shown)


pkg/pipeline/pipeline.go
```diff
@@ -114,6 +114,16 @@ func (p *Pipeline) RunForFetchedBlock(ctx context.Context, block *fetcher.Fetche
 	hasError := false
 	blockFetchTime := time.Now()
 
+	// Check if we should calculate daily rewards
+	if p.globalConfig.Rewards.CalculateRewardsDaily && !isBackfill {
```
seanmcgary, line 125: I would place this just below the `defer` function that comes after it, that way we can capture metrics if it fails

seanmcgary (approved):
Just that one comment, otherwise looks great!

PR #378, opened 2025-05-02
feat: support indexing latest blocks in real-time

### LINKED ISSUE #327: feat: support indexing blocks in real-time rather than safe blocks

serichoi65:
## Description

This PR supports indexing latest-blocks in real-time. Because these blocks can be "unsafe," reorg can happen. 
- If a reorg is detected (latest tip is behind what we've indexed), we clean up the corrupted state
- We reset our indexing position to the valid block

With these changes, you can now:
1. Index blocks in real-time with the latest (unsafe) blocks when configured for "latest" mode
2. Continue using safe blocks when configured for "safe" mode
3. Handle potential reorgs that might occur when using real-time blocks
This gives you flexibility to choose between real-time indexing (faster but with reorg risk) or safe indexing (slower but more stable) based on your needs.

To use real-time block indexing, you'll need to set the `--ethereum.latest-block-type` configuration parameter to "latest" instead of "safe". 

Fixes #327 

## Type of change

- [ ] Performance improvement

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/pkg/sidecar/blockIndexer.go
+++ b/pkg/sidecar/blockIndexer.go
@@ -64,24 +64,52 @@ func (s *Sidecar) ProcessNewBlocks(ctx context.Context) error {
 			return errors.New("Failed to get latest tip")
 		}
 
-		if latestTip < uint64(latestIndexedBlock) {
-			// If the latest tip is behind what we have indexed, this could be due to a reorg
-			// We should check if we need to revert some blocks
+		// Handle potential chain reorganization
+		if latestTip <= uint64(latestIndexedBlock) {
+			// If the latest tip is behind or equal to what we have indexed, this could be due to a reorg
 			if blockType == config.BlockType_Latest {
-				s.Logger.Sugar().Warnw("Latest tip is behind latest indexed block, possible reorg detected",
+				s.Logger.Sugar().Warnw("Latest tip is behind or equal to latest indexed block, possible reorg detected",
 					zap.Uint64("latestTip", latestTip),
 					zap.Int64("latestIndexedBlock", latestIndexedBlock))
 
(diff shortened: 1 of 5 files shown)


pkg/sidecar/blockIndexer.go
```diff
@@ -49,18 +57,36 @@ func (s *Sidecar) ProcessNewBlocks(ctx context.Context) error {
...
 		}
 
-		// If the latest tip is behind what we have indexed, sleep for a bit
 		if latestTip < uint64(latestIndexedBlock) {
-			s.Logger.Sugar().Debugw("Latest tip is behind latest indexed block, sleeping for a bit")
-			time.Sleep(BLOCK_POLL_INTERVAL)
-			continue
+			// If the latest tip is behind what we have indexed, this could be due to a reorg
+			// We should check if we need to revert some blocks
+			if blockType == config.BlockType_Latest {
```
seanmcgary, line 70: Checking if the tip is less than the latest observed block isnt quite enough to handle a re-org.

1. This could technically be a `<=` rather than just `<`. IIRC, the deepest ethereum re-org was like 2 blocks, so its entirely possible to get block `1234` and then immediately get the same block number but with a different parentHash.
2. You would want to also check to see if the block you just got from the RPC has the correct `parentHash` and then validate it going back to see if its part of the canonical chain. If the parentHash of the block you just got off chain matches the blockHash of `newBLockNumber - 1` stored in the database, then you can delete the offending block and use the new one. If it doesnt match, you'd need to trace through the DB to find where it does match and use that for pruning.

seanmcgary (changes requested):

seanmcgary (approved):

PR #386, opened 2025-05-10
fix: fixing situation where RewardsClaimed events were not parsed correctly

seanmcgary:
## Description

We ran into a log parse issue when decoding the `RewardsClaimed` event. Due to how we were parsing arguments, an unindexed log as the first argument of the event resulted in other indexed arguments not getting parsed properly.  We would end up with arguments that looked like this, missing the claimer and recipient fields:

```
"Arguments": [
    {
      "Name": "root",
      "Type": "bytes32",
      "Value": null,
      "Indexed": false
    },
    {
      "Name": "earner",
      "Type": "address",
      "Value": "0x011f63fc0f3572d7f2da01345dcf38e6a2a38c65",
      "Indexed": true
    },
    {
      "Name": "claimer",
      "Type": "address",
      "Value": "0x0427f522b0b1594eda48b49759d85e49a09d99ab",
      "Indexed": true
    },
    {
      "Name": "recipient",
      "Type": "address",
      "Value": "0x0427f522b0b1594eda48b49759d85e49a09d99ab",
      "Indexed": true
    },
    {
      "Name": "token",
      "Type": "address",
      "Value": null,
      "Indexed": false
    },
    {
      "Name": "claimedAmount",
      "Type": "uint256",
      "Value": null,
      "Indexed": false
    }
  ],
```

Turns out this was fixed in #377 but a job was never created to fix the errant transaction logs.

This PR introduces the concept of a `StartupJob`. Like a migration, StartupJobs are run once and only once but be used for shuffling data around when needed. In this case, we need to re-fetch each of the broken logs, re-parse them, update the database accordingly and re-hydrate the `rewards_claimed` table using the corrected logs.

## Type of change

## How Has This Been Tested?

Added unit tests and manually verified that this worked for all of Holesky which had ~8k logs in total for the `RewardsClaimed` event.

## Checklist:


serichoi65 (approved):
LGTM

PR #425, opened 2025-07-14
fix: remove wrapped transaction causing problems

seanmcgary:
## Description

Wrapping a query in a transaction block and then querying with a connection outside of that transaction context causes records to not actually get written to the database.

## Type of change

## Checklist:


pkg/coreContracts/helpers/helpers.go
```diff
@@ -105,21 +107,47 @@ func ImportImplementationContracts(
...
+				l.Sugar().Errorw("failed to create proxy contract",
+					zap.String("proxyAddress", i.ProxyContractAddress),
+					zap.String("implementationAddress", contract.ContractAddress),
+				)
 				return nil, err
 			}
+			if !found {
+				l.Sugar().Infow("created proxy contract",
+					zap.String("proxyAddress", i.ProxyContractAddress),
+					zap.String("implementationAddress", contract.ContractAddress),
```
serichoi65, line 143: Need to create proxy contract here

seanmcgary: Its actually created above. we're calling `FindOrCreateProxyContract` not just a find function

PR #435, opened 2025-08-06
feat: implement operatorSets rpc

serichoi65:
## Description

Implement functions for operatorSets rpc.

## Type of change

## How Has This Been Tested?

Unit tests

## Checklist:


pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,177 @@
...
+
+	query := `
+		select distinct on (staker)
+			operator
+		from staker_delegation_changes
+		where
+			staker = @staker
+			and delegated = true
+			and block_number <= @blockHeight
+		order by staker, block_number desc, log_index asc
```
seanmcgary: I dont think this is quite right. I think you actually want to do something similar to whats in the stakerDelegationSnapshots

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,177 @@
...
+	blockTimeRes := pds.db.Raw("select block_time::date from blocks where number <= @blockHeight limit 1",
+		sql.Named("blockHeight", blockHeight)).Scan(&blockTime)
+	if blockTimeRes.Error != nil {
+		return nil, blockTimeRes.Error
+	}
+
+	// Then use the block time to filter staker_operator
+	query := `
+		select distinct operator
+		from staker_operator
```
seanmcgary: Would not use this table. you're going to need to find all stakers delegated to the operator, then parse out their strategies and then dedupe the strategies

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,177 @@
...
+// ListOperatorsForAvs returns operators registered to an AVS (both registered and unregistered)
+func (pds *ProtocolDataService) ListOperatorsForAvs(ctx context.Context, avs string, blockHeight uint64) ([]OperatorSet, error) {
+	avs = strings.ToLower(avs)
+	blockHeight, err := pds.BaseDataService.GetCurrentBlockHeightIfNotPresent(ctx, blockHeight)
+	if err != nil {
+		return nil, err
+	}
+
+	query := `
+		with latest_registrations as (
```
seanmcgary: pretty sure you're going to want to union results from v1 and then results from operator_set statuses here.

when an operator registers to an operator set, they wont show up in the `avs_operator_state_changes` tables

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return operatorSets, nil
+}
+
+// ListOperatorsForBlockRange returns operators across a block range with optional filters
+// Uses staker_operator as the main table and adds WHERE clauses based on filters
+func (pds *ProtocolDataService) ListOperatorsForBlockRange(ctx context.Context, startBlock, endBlock uint64, avsAddress string, strategyAddress string, stakerAddress string) ([]string, error) {
+	// Base query using staker_operator table
+	query := `
```
seanmcgary: The real answer to this query is to capture the `OperatorRegistered` event from the delegation manager and materialize it in an EigenState model.

You could get by for now by querying the raw transactions for it though Im not sure how ill-performing that might be...

seanmcgary (changes requested):

pkg/service/protocolDataService/operators.go
serichoi65: Maybe we can use `operator_avs_strategy_snapshots` instead.

pkg/service/protocolDataService/operators.go
serichoi65: I'll make a new PR for this.

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,111 @@
...
+	"strings"
+)
+
+type OperatorSet struct {
+	Operator      string
+	OperatorSetId uint64
+}
+
+// ListOperatorsForStaker returns operators that a staker has delegated to
+func (pds *ProtocolDataService) ListOperatorsForStaker(ctx context.Context, staker string, blockHeight uint64) ([]string, error) {
```
gpsanant, line 15: i wouldn't put these under the operatorSets path, but not a big issue

serichoi65: It was already implemented as the operatorSet path in protocol-api, so I'm interested to hear about Sean's opinion on this.

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ListOperatorsForAvs returns operators registered to an AVS
+func (pds *ProtocolDataService) ListOperatorsForAvs(ctx context.Context, avs string, blockHeight uint64) ([]OperatorSet, error) {
+	avs = strings.ToLower(avs)
+	blockHeight, err := pds.BaseDataService.GetCurrentBlockHeightIfNotPresent(ctx, blockHeight)
+	if err != nil {
+		return nil, err
+	}
+
+	// First get the block time for the given block height
```
seanmcgary: Why not just get the block by `number = @blockHeight`?

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,111 @@
...
+	var blockTime string
+	blockTimeRes := pds.db.Raw("select block_time::date from blocks where number <= @blockHeight limit 1",
+		sql.Named("blockHeight", blockHeight)).Scan(&blockTime)
+	if blockTimeRes.Error != nil {
+		return nil, blockTimeRes.Error
+	}
+
+	query := `
```
seanmcgary, line 77: Can also include the block query as a CTE in this query which is cleaner IMO

pkg/service/protocolDataService/operators.go
```diff
@@ -0,0 +1,111 @@
...
```
seanmcgary, line 22: Same here as below - no need for the separate query to get the block time; just add a CTE to this query to do it

seanmcgary (changes requested):

seanmcgary (approved):

PR #465, opened 2025-10-21
feat: create ListOperatorsDelegatedStakesForStrategy endpoint

serichoi65:
## Description

- Endpoint: `GET /protocol/v1/strategies/{strategy_address}/delegated-stakes`
- Purpose: Returns delegated stakes for all operators within a specific strategy, enabling programmatic access to "top stakers" data.

## Type of change

## How Has This Been Tested?

Added unit tests for `GetTotalDelegatedOperatorSharesForStrategy`

## Checklist:


pkg/service/protocolDataService/protocol.go
```diff
@@ -152,8 +163,8 @@ func (pds *ProtocolDataService) getTotalDelegatedOperatorSharesForStrategy(ctx c
 				row_number() over (partition by staker order by block_number desc, log_index asc) as rn
 			from staker_delegation_changes
 			where
-				operator = @operator
-				and block_number <= @blockHeight
+				block_number <= @blockHeight
+				%s
```
seanmcgary: we should use the template render helper function here. using straight `sprintf` makes it hard to see what the actual query ends up being. using the template string you can include conditionals and logic directly which is much clearer

pkg/service/protocolDataService/protocol.go
```diff
@@ -196,29 +207,68 @@ func (pds *ProtocolDataService) getTotalDelegatedOperatorSharesForStrategy(ctx c
...
-	`
+		where shares > 0
+		order by shares::numeric desc;
+	`, operatorFilter)
 
-	var results struct {
-		Operator string
-		Shares   string
+	// Add pagination
+	if pagination != nil {
```
seanmcgary: for routes that have the potential to return a massive amount of data, we should set a sane default for pagination if one is not provided

seanmcgary (changes requested):

PR #466, opened 2025-10-23
feat: rewards v2.2 - Forward Looking Commitments on Unique Stake

serichoi65:
## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings


pkg/rewards/15_goldOperatorODOperatorSetRewardAmountsV2_2.go
```diff
@@ -0,0 +1,121 @@
...
+    FROM (
+        SELECT 
+            *,
+            -- We can use an arbitrary order here since the operator_tokens is the same for each (operator, strategy, hash, snapshot)
+            -- We use strategy ASC for better debuggability
+            ROW_NUMBER() OVER (
+                PARTITION BY reward_hash, snapshot, operator 
+                ORDER BY strategy ASC
+            ) AS rn
+        FROM operators_with_unique_stake
```
ypatil12: I assume this is a placeholder for now?

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,122 @@
...
+		) AS rn
+	FROM allocation_events
+),
+-- Keep only the most recent allocation for each combination
+current_allocations AS (
+	SELECT
+		operator,
+		strategy,
+		magnitude,
+		avs,
```
ypatil12, line 48: We also need the `maxMagnitude`, because the allocated stake is given by: `operatorStake * allocatedMagnitude / maxMagnitude`

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,122 @@
...
+		operator_set_id,
+		effective_block,
+		ROW_NUMBER() OVER (
+			PARTITION BY operator, strategy, avs, operator_set_id 
+			ORDER BY effective_block DESC, block_number DESC, log_index DESC
+		) AS rn
+	FROM allocation_events
+),
+-- Keep only the most recent allocation for each combination
```
ypatil12, line 42: I don't think we can keep the most recent allocation. The reason is that allocation events can be emitted for:

1. Future allocations past the current block
2. Future deallocations (via a slash). In this case if someone deallocated magnitude that was effective by block t + 14, and was slashed, then the magnitude at block t+14 would be *even* smaller. Here, we would need to use the later of the magnitudes at the effect blocks 

We also might need to think about an deallocation to one operatorSet in a block and an allocation to another (with the same stake) at the same block after the deallocation is complete. There likely needs to be a strict window

ypatil12: See rounding behavior we do in rewards V1: https://hackmd.io/u-NHKEvtQ7m7CVDb4_42bA#Part-2-Windows-amp-Snapshots

pkg/rewards/15_goldOperatorODOperatorSetRewardAmountsV2_2.go
```diff
@@ -0,0 +1,121 @@
...
+    ) t
+    -- Keep only the first row for each (operator, reward hash, snapshot)
+    WHERE rn = 1
+),
+
+-- Step 3: Calculate the tokens for each operator with dynamic split logic
```
ypatil12, line 65: The tokens of the operator should be based on the allocated stake

pkg/rewards/15_goldOperatorODOperatorSetRewardAmountsV2_2.go
```diff
@@ -0,0 +1,121 @@
+package rewards
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const _15_goldOperatorODOperatorSetRewardAmountsV2_2Query = `
```
ypatil12, line 8: Nit: I believe this isn't operator directed since the stake is being calculated based on operator allocations

pkg/rewards/16_goldStakerODOperatorSetRewardAmountsV2_2.go
```diff
@@ -0,0 +1,213 @@
...
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const _16_goldStakerODOperatorSetRewardAmountsV2_2Query = `
+CREATE TABLE {{.destTableName}} AS
+
+-- V2.2: Operator Set Rewards with UNIQUE STAKE instead of total stake
```
ypatil12, line 11: I think the easier way to calculate this is:

1. In the previous query, calculate the amount to be paid to the operator based on the operator's allocated stake
2. Get the proportion of the operator's delegated stake that each staker delegated to the operator has
3. Multiply the proportion by the amount to pay each staker

pkg/rewards/17_goldAvsODOperatorSetRewardAmountsV2_2.go
```diff
@@ -0,0 +1,265 @@
...
+
+import (
+	"database/sql"
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const _17_goldAvsODOperatorSetRewardAmountsV2_2Query = `
```
ypatil12, line 11: We should chat about refund logic. I'm not sure if we need it for this release. If we do, I think it can be simplified

ypatil12 (commented):
Preliminary pass

PR #474, opened 2025-11-14
feat: withdrawal queue schema + allocation/deallocation rounding logic

serichoi65:
### 1. Database Migration (`202511141700_withdrawalQueueAndAllocationRounding`)
- Enhanced `queued_slashing_withdrawals` table with timing columns for 14-day withdrawal tracking:
  - `completed`, `completion_*` - Completion tracking
- Enhanced `operator_allocations` table with rounding support:
  - `effective_date` - Rounded date when allocation/deallocation takes effect

### 2. Allocation/Deallocation Rounding Logic (Sabine Fork)
Implements day-boundary rounding to prevent gaming and mimic existing registration behavior:
- **Allocations (increases)**: Round **UP** to next day → starts earning tomorrow
- **Deallocations (decreases)**: Round **DOWN** to current day → stops earning immediately
- **No change**: Round **DOWN** to current day (conservative default)

### 3. Dual Behavior Implementation
Maintains backward compatibility with fork-based behavior switching:
- **Before Sabine fork**: Uses old behavior (no rounding, date fields are NULL)
- **After Sabine fork**: Applies rounding logic and populates date fields
- Model continues processing allocations from Brazos fork onwards

## Type of change

## How Has This Been Tested?

Unit tests

## Checklist:

### DIFF
--- a/pkg/eigenState/operatorAllocations/operatorAllocations.go
+++ b/pkg/eigenState/operatorAllocations/operatorAllocations.go
@@ -3,6 +3,12 @@ package operatorAllocations
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
+	"slices"
+	"sort"
+	"strings"
+	"time"
+
 	"github.com/Layr-Labs/sidecar/internal/config"
 	"github.com/Layr-Labs/sidecar/pkg/eigenState/base"
 	"github.com/Layr-Labs/sidecar/pkg/eigenState/stateManager"
@@ -10,10 +16,6 @@ import (
 	"github.com/Layr-Labs/sidecar/pkg/storage"
 	"go.uber.org/zap"
 	"gorm.io/gorm"
-	"math/big"
-	"slices"
-	"sort"
-	"strings"
 )
 
 type OperatorAllocation struct {
@@ -26,6 +28,8 @@ type OperatorAllocation struct {
 	BlockNumber     uint64
 	TransactionHash string
 	LogIndex        uint64
+	EffectiveDate   string // Rounded date when allocation takes effect (YYYY-MM-DD)
+	BlockTimestamp  string
 }
 
 type OperatorAllocationModel struct {
@@ -101,7 +105,7 @@ func (oa *OperatorAllocationModel) handleOperatorAllocationCreatedEvent(log *sto
 
 	magnitude, success := new(big.Int).SetString(outputData.Magnitude.String(), 10)
 	if !success {
-		err := fmt.Errorf("Failed to parse magnitude: %s", outputData.Magnitude.String())
+		err := fmt.Errorf("failed to parse magnitude: %s", outputData.Magnitude.String())
 		oa.logger.Sugar().Errorw("Failed to parse magnitude", zap.Error(err))
 		return nil, err
 	}
@@ -116,6 +120,28 @@ func (oa *OperatorAllocationModel) handleOperatorAllocationCreatedEvent(log *sto
 		BlockNumber:     log.BlockNumber,
 		TransactionHash: log.TransactionHash,
 		LogIndex:        log.LogIndex,
+	}
+
+	// Sabine fork: Apply rounding logic and populate date fields
+	isSabineForkActive, err := oa.IsActiveForSabineForkBlockHeight(log.BlockNumber)
+	if err != nil {
+		return nil, err
+	}
+	if isSabineForkActive {
+		effectiveDateStr, blockTimestampStr, err := oa.calculateAllocationDates(
+			log.BlockNumber,
+			magnitude,
+			strings.ToLower(outputData.Operator),
+			strings.ToLower(outputData.OperatorSet.Avs),
+			strings.ToLower(outputData.Strategy),
+			outputData.OperatorSet.Id,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		split.EffectiveDate = effectiveDateStr
+		split.BlockTimestamp = blockTimestampStr
 	}
 
 	return split, nil
@@ -338,4 +364,124 @@ func (oa *OperatorAllocationModel) IsActiveForBlockHeight(blockHeight uint64) (b
 	}
 
 	return blockHeight >= forks[config.RewardsFork_Brazos].BlockNumber, nil
+}
+
+func (oa *OperatorAllocationModel) IsActiveForSabineForkBlockHeight(blockHeight uint64) (bool, error) {
+	forks, err := oa.globalConfig.GetRewardsSqlForkDates()
+	if err != nil {
+		oa.logger.Sugar().Errorw("Failed to get rewards sql fork dates", zap.Error(err))
+		return false, err
+	}
+
+	return blockHeight >= forks[config.RewardsFork_Sabine].BlockNumber, nil
+}
+
+// calculateAllocationDates calculates the effective date and block timestamp for an allocation
+// This encapsulates the logic for querying block data and applying rounding rules
+func (oa *OperatorAllocationModel) calculateAllocationDates(
+	blockNumber uint64,
+	magnitude *big.Int,
+	operator string,
+	avs string,
+	strategy string,
+	operatorSetId uint64,
+) (effectiveDateStr string, blockTimestampStr string, err error) {
+	// 1. Get block timestamp from blocks table
+	var block storage.Block
+	result := oa.DB.Where("number = ?", blockNumber).First(&block)
+	if result.Error != nil {
+		oa.logger.Sugar().Errorw("Failed to query block timestamp",
+			zap.Error(result.Error),
+			zap.Uint64("blockNumber", blockNumber),
+		)
+		return "", "", fmt.Errorf("failed to query block timestamp: %w", result.Error)
+	}
+
+	// 2. Get previous allocation magnitude for comparison
+	previousMagnitude, err := oa.getPreviousAllocationMagnitude(
+		operator,
+		avs,
+		strategy,
+		operatorSetId,
+		blockNumber,
+	)
+	if err != nil {
+		oa.logger.Sugar().Errorw("Failed to get previous allocation magnitude",
+			zap.Error(err),
+			zap.Uint64("blockNumber", blockNumber),
+			zap.String("operator", operator),
+			zap.String("avs", avs),
+			zap.String("strategy", strategy),
+			zap.Uint64("operatorSetId", operatorSetId),
+		)
+		return "", "", fmt.Errorf("failed to get previous allocation magnitude: %w", err)
+	}
+
+	// 3. Determine effective date using rounding rules
+	effectiveDate := oa.determineEffectiveDate(block.BlockTime, magnitude, previousMagnitude)
+
+	// 4. Format and return date strings
+	effectiveDateStr = effectiveDate.Format("2006-01-02")
+	blockTimestampStr = block.BlockTime.Format(time.RFC3339)
+
+	return effectiveDateStr, blockTimestampStr, nil
+}
+
+// getPreviousAllocationMagnitude retrieves the most recent allocation magnitude
+// for the given operator-avs-strategy combination before the specified block number
+func (oa *OperatorAllocationModel) getPreviousAllocationMagnitude(
+	operator string,
+	avs string,
+	strategy string,
+	operatorSetId uint64,
+	currentBlockNumber uint64,
+) (*big.Int, error) {
+	var previousAllocation OperatorAllocation
+
+	result := oa.DB.
+		Where("operator = ?", operator).
+		Where("avs = ?", avs).
+		Where("strategy = ?", strategy).
+		Where("operator_set_id = ?", operatorSetId).
+		Where("block_number < ?", currentBlockNumber).
+		Order("block_number DESC, log_index DESC").
+		Limit(1).
+		First(&previousAllocation)
+
+	if result.Error != nil {
+		if result.Error == gorm.ErrRecordNotFound {
+			// No previous allocation found, return 0
+			return big.NewInt(0), nil
+		}
+		return nil, result.Error
+	}
+
+	magnitude, success := new(big.Int).SetString(previousAllocation.Magnitude, 10)
+	if !success {
+		return nil, fmt.Errorf("failed to parse previous magnitude: %s", previousAllocation.Magnitude)
+	}
+
+	return magnitude, nil
+}
+
+// determineEffectiveDate determines the effective date for an allocation based on rounding rules
+// - Allocation (increase): Round UP to next day
+// - Deallocation (decrease): Round DOWN to current day
+func (oa *OperatorAllocationModel) determineEffectiveDate(
+	blockTimestamp time.Time,
+	newMagnitude *big.Int,
+	previousMagnitude *big.Int,
+) time.Time {
+	blockTimestamp = blockTimestamp.UTC()
+	comparison := newMagnitude.Cmp(previousMagnitude)
+
+	year, month, day := blockTimestamp.Date()
+	midnightUTC := time.Date(year, month, day, 0, 0, 0, 0, time.UTC)
+
+	if comparison > 0 {
+		// Allocation (increase) - always round up to next day
+		return midnightUTC.Add(24 * time.Hour)
+	}
+	// Deallocation (decrease or no change) - round down to current day
+	return midnightUTC
 }
--- a/pkg/eigenState/operatorAllocations/operatorAllocations.go
+++ b/pkg/eigenState/operatorAllocations/operatorAllocations.go
@@ -29,7 +29,6 @@ type OperatorAllocation struct {
 	TransactionHash string
 	LogIndex        uint64
 	EffectiveDate   string // Rounded date when allocation takes effect (YYYY-MM-DD)
-	BlockTimestamp  string
 }
 
 type OperatorAllocationModel struct {
@@ -128,7 +127,7 @@ func (oa *OperatorAllocationModel) handleOperatorAllocationCreatedEvent(log *sto
 		return nil, err
 	}
 	if isSabineForkActive {
-		effectiveDateStr, blockTimestampStr, err := oa.calculateAllocationDates(
+		effectiveDateStr, err := oa.calculateAllocationDates(
 			log.BlockNumber,
 			magnitude,
 			strings.ToLower(outputData.Operator),
@@ -141,7 +140,6 @@ func (oa *OperatorAllocationModel) handleOperatorAllocationCreatedEvent(log *sto
 		}
 
 		split.EffectiveDate = effectiveDateStr
-		split.BlockTimestamp = blockTimestampStr
 	}
 
 	return split, nil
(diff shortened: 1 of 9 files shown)


pkg/eigenState/operatorAllocations/operatorAllocations.go
```diff
@@ -339,3 +363,123 @@ func (oa *OperatorAllocationModel) IsActiveForBlockHeight(blockHeight uint64) (b
...
+	newMagnitude *big.Int,
+	previousMagnitude *big.Int,
+) time.Time {
+	blockTimestamp = blockTimestamp.UTC()
+	comparison := newMagnitude.Cmp(previousMagnitude)
+
+	year, month, day := blockTimestamp.Date()
+	midnightUTC := time.Date(year, month, day, 0, 0, 0, 0, time.UTC)
+
+	if comparison > 0 {
```
ypatil12: Do we need to round up on the API? My thinking here is that the API can serve the allocation at the block second granularity.

ypatil12: From my understanding, we don't need to return rounded up values on staker shares/operaror shares via the API

pkg/eigenState/operatorAllocations/operatorAllocations.go
```diff
@@ -339,3 +363,123 @@ func (oa *OperatorAllocationModel) IsActiveForBlockHeight(blockHeight uint64) (b
...
+	effectiveDateStr = effectiveDate.Format("2006-01-02")
+
+	return effectiveDateStr, nil
+}
+
+// getPreviousAllocationMagnitude retrieves the most recent allocation magnitude
+// for the given operator-avs-strategy combination before the specified block number
+func (oa *OperatorAllocationModel) getPreviousAllocationMagnitude(
+	operator string,
+	avs string,
```
ypatil12: Nit: Generally makes sense to put operatorSetID after avs

pkg/eigenState/operatorAllocations/operatorAllocations.go
```diff
@@ -339,3 +363,123 @@ func (oa *OperatorAllocationModel) IsActiveForBlockHeight(blockHeight uint64) (b
...
+	operatorSetId uint64,
+	currentBlockNumber uint64,
+) (*big.Int, error) {
+	var previousAllocation OperatorAllocation
+
+	result := oa.DB.
+		Where("operator = ?", operator).
+		Where("avs = ?", avs).
+		Where("strategy = ?", strategy).
+		Where("operator_set_id = ?", operatorSetId).
```
ypatil12: operatorSetId should be after AVS

pkg/postgres/migrations/202511141700_withdrawalQueueAndAllocationRounding/up.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (m *Migration) Up(db *sql.DB, grm *gorm.DB, cfg *config.Config) error {
+	queries := []string{
+		// =============================================================================
+		// PART 1: Enhance queued_slashing_withdrawals to support withdrawal queue
+		// =============================================================================
+
+		// Add completion tracking (timestamps can be derived from blocks table via FK)
+		`alter table queued_slashing_withdrawals add column if not exists completed boolean default false`,
```
ypatil12: Don't need the completed boolean? Just the block number at which it can be completed? Or, has the thinking here changed?

pkg/postgres/migrations/202511141700_withdrawalQueueAndAllocationRounding/up.go
```diff
@@ -0,0 +1,61 @@
...
+		`create index if not exists idx_queued_withdrawals_completed on queued_slashing_withdrawals(completion_block_number) where completed = true`,
+
+		// =============================================================================
+		// PART 3: Update operator_allocations table for rounding logic
+		// =============================================================================
+
+		// Add effective_date column for allocation/deallocation rounding
+		// This is a computed value based on magnitude changes (round UP for increases, DOWN for decreases)
+		// block_timestamp can be derived from block_number FK to blocks table
+		`alter table operator_allocations add column if not exists effective_date date`,
```
ypatil12: Rounding happens during the rewards calculation itself. See `stakerShareSnapshots.go` or `stakerDelegationSnapshots.go`

ypatil12: I believe we want to create a new table instead of altering the current one?

ypatil12: Eg:

```sql
		`CREATE TABLE IF NOT EXISTS staker_share_snapshots (
			staker varchar not null,
			strategy varchar not null,
			shares numeric not null,
			snapshot date not null
		)
		`,
		`create index idx_staker_share_snapshots_staker_strategy_snapshot on staker_share_snapshots (staker, strategy, snapshot)`,
		`create index idx_staker_share_snapshots_strategy_snapshot on staker_share_snapshots (strategy, snapshot)`,
	}
```

ypatil12: `effective_date` should be `snapshot_date`?

pkg/eigenState/operatorAllocations/operatorAllocations.go
ypatil12: Rounding should be done during the rewards calculation, specifically in the snapshots tables

serichoi65: Removed all rounding logic from operatorAllocations.go and added rounding logic in the snapshot tables.

pkg/postgres/migrations/202511141700_withdrawalQueueAndAllocationRounding/up.go
serichoi65: No longer altering `operator_allocations`

pkg/postgres/migrations/202511141700_withdrawalQueueAndAllocationRounding/up.go
serichoi65: Ah yes, we do not need completed boolean. Removed.

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,118 @@
...
+			strategy,
+			operator_set_id,
+			magnitude,
+			block_time,
+			LAG(magnitude) OVER (PARTITION BY operator, avs, strategy, operator_set_id ORDER BY block_time, block_number, log_index) as previous_magnitude,
+			-- Allocation (increase): Round UP to next day
+			-- Deallocation (decrease or no change): Round DOWN to current day
+			CASE
+				WHEN LAG(magnitude) OVER (PARTITION BY operator, avs, strategy, operator_set_id ORDER BY block_time, block_number, log_index) IS NULL THEN
+					-- First allocation: round down to current day (conservative default)
```
ypatil12: The comment here says round down and it looks like  but the tests do what we want (round up). However, this code looks like we round down on the first allocation, when we should round up

ypatil12: So basically, first allocation & subsequent increases round up. Deallocations round down.

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,118 @@
...
+	),
+	-- Get the latest record for each day
+	daily_records as (
+		SELECT
+			operator,
+			avs,
+			strategy,
+			operator_set_id,
+			magnitude,
+			block_time,
```
ypatil12, line 25: So the `AllocationUpdated` event emits an `effectBlock`. I'm not sure if we're using that anywhere here. The `effectBlock` is the block at which the allocation/dealloction becomes active

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,145 @@
...
+			WHERE operator = ? AND avs = ? AND strategy = ? AND operator_set_id = ?
+			ORDER BY snapshot
+		`, "0xoperator1", "0xavs1", "0xstrategy1", 1).Scan(&snapshots)
+		assert.Nil(t, res.Error)
+		assert.True(t, len(snapshots) > 0, "Expected snapshots to be generated")
+
+		// Verify rounding behavior:
+		// - First allocation (1000) at block 100 (2024-11-14 10:00) should have snapshot starting 2024-11-15
+		// - Increase (1500) at block 101 (2024-11-14 15:00) should have snapshot starting 2024-11-15 (next day after 2024-11-14)
+		// - Decrease (500) at block 102 (2024-11-15 12:00) should have snapshot starting 2024-11-15 (same day)
```
ypatil12: An edge case to think about on testnet is that since the allocations/deallocations can happen on the order of minutes, if you round deallocations down you may have the following?

1. 12/1 @ 12pm: Allocate 500
2. 12/1 @ 1pm: Deallocate down to 400

The behavior here should be that the magnitude for the day is 400

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+		// Find the snapshot with magnitude 1000
+		var mag1000Count int64
+		res = grm.Raw(`
+			SELECT COUNT(*) FROM operator_allocation_snapshots
+			WHERE operator = ? AND avs = ? AND strategy = ? AND operator_set_id = ?
+			AND magnitude = '1000' AND snapshot = ?
```
ypatil12, line 293: Shouldn't the magnitude here be 1500 on 11/15? This is because we had the initially allocation then an increase on the same day. we took the latest record of the day, so that's 1500

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,145 @@
...
+		`, "0xoperator1", "0xavs1", "0xstrategy1", 1, "2024-11-15").Scan(&mag1000Count)
+		assert.Nil(t, res.Error)
+		assert.True(t, mag1000Count > 0, "Expected magnitude 1000 to have snapshot on 2024-11-15 (rounded up)")
+	})
```
ypatil12, line 406: We should add the following tests:

1. Allocate, check round up
2. Allocate, allocate again on same day. Use the latest allocation
3. Allocate, then deallocate on same day. This is possible on testnet.  Use the latest deallocation value
4. Allocate on day 1 @ 12pm. Deallocate down to 0 on day 2 @12pm. Should see no days counted
5. Allocate on day 1 @12pm. Deallocate down to 0 on day 3 @12pm. Should have 1 day counted

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: Fixed to first allocation to round UP

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: Changed join from `oa.block_number` to `oa.effective_block`

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,118 @@
...
+	"go.uber.org/zap"
+)
+
+const operatorAllocationSnapshotsQuery = `
+	insert into operator_allocation_snapshots(operator, avs, strategy, operator_set_id, magnitude, snapshot)
+	WITH ranked_allocation_records as (
+		SELECT *,
+			   ROW_NUMBER() OVER (PARTITION BY operator, avs, strategy, operator_set_id, cast(block_time AS DATE) ORDER BY block_time DESC, log_index DESC) AS rn
+		FROM operator_allocations oa
+		INNER JOIN blocks b ON oa.effective_block = b.number
```
ypatil12, line 14: There can be allocations that are in the future. For example, take the following scenario:

Block 1, Timestamp 1: Alice allocates 5 magnitude. This is effective on block 10

Sidecar then processes this event. Since we do an inner the allocation is discarded. However, upon the block 10 actually existing the allocation is then processed. 

That is to say, I think we need to add a test for this scenario :)

pkg/rewards/operatorAllocationSnapshots.go
ypatil12: We're joining on the `effective_block`, but now we also need to make sure that we're making sure that the allocation is counted at the effective block and its associated day, not the block at which the event was emitted. 

Example is when the event is emitted at block 1 but the allocation is activated at block 10

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,398 @@
...
+
+		// Create test operator allocations
+		// First allocation: magnitude 1000 at block 100 (should round up to 2024-11-15)
+		res := grm.Exec(`
+			INSERT INTO operator_allocations (operator, avs, strategy, operator_set_id, magnitude, effective_block, block_number, transaction_hash, log_index, created_at, updated_at)
+			VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, NOW(), NOW())
+		`, "0xoperator1", "0xavs1", "0xstrategy1", 1, "1000", 100, 100, "tx_100", 1)
+		assert.Nil(t, res.Error)
+
+		// Increase allocation: magnitude 1500 at block 101 (should round up to 2024-11-16)
```
ypatil12, line 94: Shouldn't this round up to 2024-11-15 since block 101 is at 2024-11-14?

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,398 @@
...
+			ORDER BY snapshot
+		`, "0xoperator1", "0xavs1", "0xstrategy1", 1).Scan(&snapshots)
+		assert.Nil(t, res.Error)
+		assert.True(t, len(snapshots) > 0, "Expected snapshots to be generated")
+
+		// Verify rounding behavior:
+		// - Both block 100 (magnitude 1000) and block 101 (magnitude 1500) occur on 2024-11-14
+		// - SQL takes the latest record per day (block 101 with magnitude 1500)
+		// - This allocation is an increase, so it rounds up to 2024-11-15
+		// - Decrease (500) at block 102 (2024-11-15 12:00) rounds down to 2024-11-15 (same day)
```
ypatil12: A bit confused here, since we round up 2 allocations and round down 1 deallocation. The deallocation was the latest event that occurred and therefore the allocation should be at 500.

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,118 @@
+package rewards
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const operatorAllocationSnapshotsQuery = `
```
ypatil12, line 8: Taking a look at `operatorAVSRegistrationSnapshots.go`, I think we might be missing a few cases in general:

From that file:

```golang
// Operator AVS Registration Windows: Ranges at which an operator has registered for an AVS
// 0. Ranked: Rank the operator state changes by block_time and log_index since sqlite lacks LEAD/LAG functions
// 1. Marked_statuses: Denote which registration status comes after one another
// 2. Removed_same_day_deregistrations: Remove a pairs of (registration, deregistration) that happen on the same day
// 3. Registration_periods: Combine registration together, only select registrations with:
// a. (Registered, Unregistered)
// b. (Registered, Null). If null, the end time is the current timestamp
// 4. Registration_snapshots: Round up each start_time to  0 UTC on NEXT DAY and round down each end_time to 0 UTC on CURRENT DAY
// 5. Operator_avs_registration_windows: Ranges that start and end on same day are invalid
// Note: We cannot assume that the operator is registered for avs at end_time because it is
// Payments calculations should only be done on snapshots from the PREVIOUS DAY. For example say we have the following:
// <-----0-------1-------2------3------>
// ^           ^
// Entry        Exit
// Since exits (deregistrations) are rounded down, we must only look at the day 2 snapshot on a pipeline run on day 3.
```

ypatil12: 1. If an allocation and a non-zero deallocation happens on the same day (this is just on testnet), we should just use the deallocation and not round down. Rounding down would just give the user an extra day of stake. 
2. If an allocation and *zero* deallocation happens on the same day, then discard the entire allocation/deallocation pair

Again, these cases only exist on testnet since allocation/deallocation can happen in the same day. In practice mainnet would never have this problem.

ypatil12: Actually it looks like we properly handle these cases based on the tests. I just want to leave this comment up to make sure you agree that we're not missing anything based on the what we did in the above logic too :)

ypatil12 (commented):
One minor comment on the `effective_block` and a test to add. Other than looking solid!

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: yes, tests properly handle allocation/deallocation on same day by taking latest record!

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: correct, 
```
INNER JOIN blocks b ON oa.effective_block = b.number
WHERE b.block_time < TIMESTAMP '{{.cutoffDate}}'
```
The join is on `effective_block`, so b.block_time comes from the effective block, not the emission block. The new test case added for the future allocation scenario verifies this

pkg/rewards/operatorAllocationSnapshots_test.go
```diff
@@ -0,0 +1,443 @@
...
+		// - Since 1500 > 1000 (increase), rounds up to 2024-11-15
+		// - Block 102 (2024-11-15 12:00, mag 500): decrease from 1500 to 500, rounds down to 2024-11-15
+		// - Final allocation: 500 (the latest/current magnitude after all operations)
+
+		// Find the snapshot with magnitude 1500 (latest allocation on 2024-11-14)
+		var mag1500Count int64
+		res = grm.Raw(`
+			SELECT COUNT(*) FROM operator_allocation_snapshots
+			WHERE operator = ? AND avs = ? AND strategy = ? AND operator_set_id = ?
+			AND magnitude = '1500' AND snapshot = ?
```
ypatil12: Shouldn't the final magnitude be 500?

ypatil12: 11.14: 2 allocation. Once to 1000. Another to 1500. This is round dup to 11.15

11.15. One deallocation down to 500. This is rounded down to 11.15. Thus, we should have a magnitude of 500

Maybe there is an edge case here?

serichoi65: Yes, it should be 500. 

When an allocation increase (1500 on 11/14) rounds UP to 2024-11-15, and a deallocation (500 on 11/15) rounds DOWN to the same date, the SQL's `LEAD()` function causes the earlier event's time window to become [2024-11-15, 2024-11-15). This zero-length window gets filtered out in the cleaned_records CTE, leaving only the chronologically later event (magnitude 500) in the final snapshot.

ypatil12: Nice! See that the tests reflects this :)

ypatil12:
This LGTM!

PR #476, opened 2025-11-17
feat: withdrawalQueue and deallocationQueue logic into rewards calculation

serichoi65:
### Problem
Operators were not earning rewards during the deallocation queue period, even though their allocations remain at risk until the `effective_date`.

**Current behavior:**
- Operator deallocates 100 from allocation on Day 1 (effective_date = Day 15)
- `operator_allocations` table immediately shows reduced allocation
- Rewards calculation uses reduced amount → ❌ Operator loses 14 days of rewards

**Expected behavior:**
- Operator should continue earning on OLD (higher) allocation until `effective_date`
- After `effective_date`, earn on NEW (lower) allocation

### Solution
**Deallocation Queue Adjustment**:
Added `AdjustOperatorShareSnapshotsForDeallocationQueue()` function that:
1. Reads `magnitude_decrease` from `deallocation_queue_snapshots`
2. Aggregates across all AVSs per (operator, strategy)
3. Adds the decrease back to `operator_share_snapshots` during queue period

This mirrors the existing withdrawal queue adjustment for stakers.

**Operator Allocation Snapshots**:
New operatorAllocationSnapshots.go generates daily snapshots with allocation rounding:

- Allocations round UP to next day (allocation @ 5pm on 1/4 → starts 1/5)
- Deallocations round DOWN to previous day (deallocation @ 5pm on 1/4, effective 1/18 → reduced allocation starts 1/17)
- Tracks max_magnitude for slashable stake calculations

**Gold Tables 15-17 (Unique Stake Rewards):**

- Table 15: Operator rewards based on allocated magnitude
- Table 16: Staker rewards (distributed from Table 15)
- Table 17: AVS refunds for unallocated periods

### Changes

- Added deallocationQueueShareSnapshots.go - deallocation queue adjustment
- Added operatorAllocationSnapshots.go - allocation snapshot generation with rounding
- Added Gold Tables 15-17 for unique stake reward distribution
- Integrated into rewards calculation pipeline in rewards.go
- Migration: 202511141700_withdrawalQueueAndAllocationRounding - creates operator_allocation_snapshots table

## Type of change

## Checklist:


pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+// - Withdrawals are queued when SlashingWithdrawalQueued event occurs
+// - Shares become withdrawable after 14 days (withdrawable_date)
+// - Until withdrawable_date, shares should still earn rewards
+// - After withdrawable_date, staker can withdraw at any time (no longer forced risk)
+//
+// For a given snapshot_date, we add back shares where:
+// - queued_date <= snapshot_date (withdrawal was already queued)
+// - withdrawable_date > snapshot_date (shares still in forced queue period)
+//
+// NOTE: We do NOT check 'completed = false' because:
```
ypatil12: What's the point of even storing `completed`?

ypatil12: Is this for the API? When a withdrawal is completed do we index and update the completed field?

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,198 @@
...
+			(prev_magnitude::numeric - new_magnitude::numeric) as magnitude_decrease,
+			effective_date,
+			block_date
+		from deallocation_adjustments
+		where
+			-- Only deallocations (decreases)
+			prev_magnitude is not null
+			and new_magnitude::numeric < prev_magnitude::numeric
+	)
+	insert into deallocation_queue_snapshots(
```
ypatil12: Are we not inserting allocations into the deallocation queue snapshots? How do we handle an deallocation that occurs immediately after an allocation?

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,198 @@
...
+	)
+	select
+		operator,
+		avs,
+		strategy,
+		magnitude_decrease,
+		block_date,
+		effective_date,
+		'{{.snapshotDate}}'::date as snapshot
+	from deallocations_only
```
ypatil12: Don't we need to have operatorSetId?

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+// 1. Withdrawal can be completed anytime after withdrawable_date
+// 2. Once withdrawable_date passes, staker is no longer forced to take risk
+// 3. So rewards should stop at withdrawable_date, not completion_date
+//
+// The query joins with blocks table to derive timestamps from block_number.
+const withdrawalQueueShareSnapshotsQuery = `
```
ypatil12: How does this handle slashings in the queue?

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+		snapshot
+	)
+	select
+		staker,
+		strategy,
+		shares,
+		queued_date,
+		withdrawable_date,
+		'{{.snapshotDate}}'::date as snapshot
+	from withdrawal_queue_adjustments
```
ypatil12: Where are we re-inserting the shares into the `staker_shares` table once the withdrawal is completable?

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+//
+// The query joins with blocks table to derive timestamps from block_number.
+const withdrawalQueueShareSnapshotsQuery = `
+	with withdrawal_queue_adjustments as (
+		select
+			qsw.staker,
+			qsw.strategy,
+			qsw.shares_to_withdraw as shares,
+			b_queued.block_time as queued_timestamp,
+			date(b_queued.block_time) as queued_date,
```
ypatil12: Why is rounding being done here (cast block_time to date)? My thinking is that the purpose of this table is to be used to keep track of shares that were in the queue and then once they are completable we insert a diff back into the `staker_shares` table.

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,198 @@
...
+			oa.operator,
+			oa.avs,
+			oa.strategy,
+			oa.magnitude as new_magnitude,
+			oa.effective_date,
+			date(b.block_time) as block_date,
+			oa.block_number,
+			-- Get previous allocation to calculate the difference
+			lag(oa.magnitude) over (
+				partition by oa.operator, oa.avs, oa.strategy
```
ypatil12: Needs to also use the operatorSetID

pkg/rewards/deallocationQueueShareSnapshots.go
ypatil12: Nit: rest of snapshots query have the insert into in the beginning

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,198 @@
...
+		where
+			-- Only deallocations (decreases)
+			prev_magnitude is not null
+			and new_magnitude::numeric < prev_magnitude::numeric
+	)
+	insert into deallocation_queue_snapshots(
+		operator,
+		avs,
+		strategy,
+		magnitude_decrease,
```
ypatil12: Why store the diff? Fine to just store the magnitude itself and then get the latest magnitude for the given day upon calculation of the slashable stake

ypatil12:
The withdrawals are still processed immediately instead of 14 days after. The [stakerShares](https://github.com/Layr-Labs/sidecar/blob/master/pkg/eigenState/stakerShares/stakerShares.go) file also has to be edited to reflect this. One thing to think about here is that we would (presumably) want the sidecar API to return the actual shares of the user in the core protocol, which should immediately decremented on a queue. However, for rewards we do not want the shares to be immediately decremented.

pkg/rewards/withdrawalQueueShareSnapshots.go
serichoi65: It is an extra field I thought could be useful, but we can remove for now. If an api needs it, I can add it then.

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: The query only filters for decreases only:
```
where prev_magnitude is not null
  and new_magnitude::numeric < prev_magnitude::numeric
```

For a deallocation immediately after an allocation:

- The allocation creates a new magnitude with its own effective_date (rounds UP)
- The subsequent deallocation would be captured with LAG() seeing the new (higher) magnitude as prev_magnitude
- The decrease amount would be calculated correctly

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: yes, this is a bug! Added.

pkg/rewards/withdrawalQueueShareSnapshots.go
serichoi65: The flow:
1. Staker has 100 shares
2. Withdrawal queued → creates -100 delta (shares now 0)
3. During 14-day queue: withdrawal queue adds back +100 for rewards
4. IF slashing occurs during queue → slashing creates additional negative delta
5. Net result: Staker earns rewards on (queued shares - slashed amount)

Slashings are applied at the `staker_share_deltas` level before the queue adjustment, so the math works out correctly!

pkg/rewards/withdrawalQueueShareSnapshots.go
serichoi65: `staker_shares` represents cumulative shares (mirrors protocol state), while `staker_share_snapshots` captures daily snapshots used specifically for rewards calculation.

When a withdrawal is queued:

1. A negative delta is created in `staker_share_deltas`
2. `staker_shares` will show reduced shares
3. During queue period: AdjustStakerShareSnapshotsForWithdrawalQueue adds shares back to staker_share_snapshots for rewards
4. After withdrawable_date: Shares are NOT added back → rewards stop

pkg/rewards/withdrawalQueueShareSnapshots.go
serichoi65: This is date rounding, not allocation rounding. This is converting block timestamps to dates to figure out what day a withdrawal was queued, compare with snapshot_date, and calculating +14 day withdrawable_date.

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: It would also work to store prev_magnitude, but I think it's easier to track the diff for multiple pending deallocations and easier aggregation.

serichoi65:
@ypatil12 This is the current architecture:
```
Blockchain Event (SlashingWithdrawalQueued)
  ↓
eigenState: Create negative delta (protocol truth)
  ↓
staker_share_deltas table (single source of truth)
  ↓
Rewards: Read deltas + apply queue adjustment
```

- eigenState should always reflect protocol state
  - When SlashingWithdrawalQueued fires, shares ARE immediately decremented on-chain
  - eigenState indexer mirrors this exactly - that's its job
  - Changing this breaks the "state mirror" principle
- Single source of truth
  - All downstream consumers read from staker_share_deltas
  - Rewards calculator applies its domain logic on top
  - Clean data flow: raw state → business logic
 
That's why it would be better not to edit stakerShares but having additional deltas table to correctly calculate rewards.

pkg/rewards/withdrawalQueueShareSnapshots.go
ypatil12: Yeah, makes sense for API, not for rewards calculation

pkg/rewards/deallocationQueueShareSnapshots.go
ypatil12: One invariant of the core protocol is there can be at most 1 pending allocation or deallocation at a time. Unsure if we can take advantage of this. It seems like we're not using the diff in the `operatorAllocationSnapshots` file

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,203 @@
...
+//
+// For a given snapshot_date, we need to identify deallocations where:
+// - block_date <= snapshot_date (deallocation already recorded)
+// - effective_date > snapshot_date (not yet effective)
+//
+// This is conceptually similar to withdrawal queue but uses effective_date instead of
+// a fixed 14-day period.
+//
+// NOTE: This handles DECREASES (deallocations). Increases (allocations) are simpler
+// because they round UP, so they naturally start earning on effective_date.
```
ypatil12: Im confused why we have a `operatorAllocationSnapshots` table and a `deallocationQueueShareSnapshots` table? Is it possible to have a single table with allocation/deallocation updates. When a deallocation event occurs, insert it into the table and only process at the effect block

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,203 @@
...
+		zap.String("snapshotDate", snapshotDate),
+		zap.Int64("rowsAffected", res.RowsAffected),
+	)
+
+	return nil
+}
+
+func (r *RewardsCalculator) AdjustOperatorShareSnapshotsForDeallocationQueue(snapshotDate string) error {
+	adjustQuery := `
+	insert into operator_share_snapshots(operator, strategy, shares, snapshot)
```
ypatil12: Confused why we're adding magnitudes (percentages) into the operator shares? Did you mean to add into `operatorAllocationSnapshots` ?

pkg/rewards/deallocationQueueShareSnapshots.go
```diff
@@ -0,0 +1,203 @@
...
+		select
+			dqs.operator,
+			dqs.strategy,
+			sum(dqs.magnitude_decrease::numeric)::text as shares,
```
ypatil12: The magnitude decrease isn't the shares. The shares of an operator are given by a different set of events: `OperatorSharesIncreased` and `OperatorSharesDecreased`.

pkg/rewards/deallocationQueueShareSnapshots.go
ypatil12: Isn't the above logic handled in `OperatorAllocationSnapshots` in #474 , confused what the `deallocation_queue_snapshots` is doing

pkg/rewards/withdrawalQueueShareSnapshots.go
ypatil12: I don't think we need to do date rounding either. Ideally we just insert the share difference, once the withdrawal is completable, into the `staker_shares` table. That way, we let the logic we've already written handle the date rounding (`staker_share_snapshots`)

pkg/rewards/rewards.go
```diff
@@ -673,12 +681,31 @@ func (rc *RewardsCalculator) generateSnapshotData(snapshotDate string) error {
...
 	rc.logger.Sugar().Debugw("Generated operator allocation snapshots")
 
+	// Generate deallocation queue snapshots - operators continue earning on old allocation until effective_date
+	if err = rc.GenerateAndInsertDeallocationQueueSnapshots(snapshotDate); err != nil {
+		rc.logger.Sugar().Errorw("Failed to generate deallocation queue snapshots", "error", err)
+		return err
+	}
+	rc.logger.Sugar().Debugw("Generated deallocation queue snapshots")
+
+	if err = rc.AdjustOperatorShareSnapshotsForDeallocationQueue(snapshotDate); err != nil {
```
ypatil12: This shouldn't have to happen since operator shares (think total stake) are orthogonal to the deallocation queue. The *slashable* shares of the operator do relate and those are properly calculated in the gold tables themselves.

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+		zap.String("snapshotDate", snapshotDate),
+		zap.Int64("rowsAffected", res.RowsAffected),
+	)
+
+	return nil
+}
+
+func (r *RewardsCalculator) AdjustStakerShareSnapshotsForWithdrawalQueue(snapshotDate string) error {
+	adjustQuery := `
+	insert into staker_share_snapshots(staker, strategy, shares, snapshot)
```
ypatil12: I'm not sure if we can actually do this. My hunch is that we can run into issues with the snapshotting that's already been done and overwrite state. For example, say the staker shares table looked like:

Day 0: Alice queues a withdrawal for 30 shares. 

1. Day 1: Alice has 35 shares
2. Day 2: Alice has 40 shares (Deposits 5 extra)

Then on Day 2, we also apply a queued withdrawal decrementing her shares down to 10. How do we handle this diff appropriately? 

Furthermore, `staker_shares` keeps track of a *running sum* of shares and am not sure this handles the shares that were previously credit to the staker.

pkg/rewards/withdrawalQueueShareSnapshots.go
ypatil12: A design principle that makes sense is that we should do as *little* rounding as possible. For this functionality, I'd like us to ideally just insert into the `staker_shares` table. With that in mind, maybe we should have a parallel staker shares table just for rewards?

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,170 @@
...
+
+	r.logger.Sugar().Infow("Generated withdrawal queue share snapshots",
+		zap.String("snapshotDate", snapshotDate),
+		zap.Int64("rowsAffected", res.RowsAffected),
+	)
+
+	return nil
+}
+
+func (r *RewardsCalculator) AdjustStakerShareSnapshotsForWithdrawalQueue(snapshotDate string) error {
```
ypatil12: Which function is used to insert shares back into staker share snapshots and which is used to decrement once the withdrawal is completable?

ypatil12 (commented):
I think it would be helpful to have sequence diagrams on how events are processed, including the propogtaion of an event through each file, for allocation/deallocation and withdrawals

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: Ah yes, I added the `deallocationQueueShareSnapshots` before adding `operatorAllocationSnapshots` but it is a good idea to have as a single table. Updated.

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: This is a mistake :/ Fixed.

pkg/rewards/deallocationQueueShareSnapshots.go
serichoi65: Removed

pkg/rewards/rewards.go
serichoi65: Removed

pkg/rewards/withdrawalQueueShareSnapshots.go
serichoi65: - Insert back: `AdjustStakerShareSnapshotsForWithdrawalQueue()` adds queued shares during 14-day window
- Decrement: Automatically handled! After `withdrawable_date` (Day 14+), the query doesn't match anymore, so shares stop being added back

pkg/postgres/migrations/202511171438_withdrawalAndDeallocationQueues/up.go
```diff
@@ -0,0 +1,45 @@
...
+			staker varchar not null,
+			strategy varchar not null,
+			shares numeric not null,
+			queued_date date not null,
+			withdrawable_date date not null,
+			snapshot date not null,
+			primary key (staker, strategy, snapshot),
+			constraint uniq_withdrawal_queue_share_snapshots unique (staker, strategy, snapshot)
+		)`,
+		`create index if not exists idx_withdrawal_queue_share_snapshots_snapshot on withdrawal_queue_share_snapshots(snapshot)`,
```
ypatil12: I don't think the withdrawal queue shares need to be snapshotted (have daily rounding). We should just insert them into staker shares 

My thinking here is that:

1. Alice queue a withdrawal. The staker shares table has a decrement for 100 shares
2. We add back 100 shares into the staker shares table to cancel out the withdrawal
3. 14 days later, we add a decrement of 100 shares into the staker shares table so that its processed properly

ypatil12: Unless the above is overly complex? But I do think it is clean to only snapshot on top of the staker shares table and not withdrawals.

ypatil12: The other option is for rewards, have an if case where we don't add withdrawals immediately to staker shares. We just add it to withdrawal_queue table. Then, we do step 3 above. Likely is cleanest

pkg/rewards/withdrawalQueueShareSnapshots.go
```diff
@@ -0,0 +1,117 @@
...
+		return res.Error
+	}
+
+	if maxBlock < sabineFork.BlockNumber {
+		r.logger.Sugar().Debugw("Snapshot date is before Sabine fork, skipping withdrawal queue logic",
+			zap.String("snapshotDate", snapshotDate),
+			zap.Uint64("maxBlock", maxBlock),
+			zap.Uint64("sabineForkBlock", sabineFork.BlockNumber),
+		)
+		return nil
```
ypatil12: Think we should insert this into staker_shares table

ypatil12 (commented):
I think it might be worth writing down a diagram or flow on how the staker shares propagate based on a withdrawal and the tables at each step of the process

pkg/rewards/withdrawalQueueShareSnapshots.go
ypatil12: IMO it's best for us to rely on the snapshotting only done on the `staker_shares` table. We simply need to be smart about *when* to insert a withdrawal into that table

ypatil12: So, when a withdrawal happens, we don't insert the share decrement. After a withdrawal is completable we do insert the share decrement

pkg/postgres/migrations/202511171438_withdrawalAndDeallocationQueues/up.go
seanmcgary: Rather than thrashing the table and having to subtract, then add, then subtract again, why not create a new table to record these cases that you can then sum together with the staker shares table?

pkg/postgres/migrations/202512021500_stakerSharesView/up.go
```diff
@@ -0,0 +1,42 @@
...
+	"gorm.io/gorm"
+)
+
+type Migration struct {
+}
+
+func (m *Migration) Up(db *sql.DB, grm *gorm.DB, cfg *config.Config) error {
+	queries := []string{
+		// VIEW combines base shares + withdrawal queue adjustments
+		`create or replace view staker_share_snapshots_final as
```
seanmcgary: creating this as a view is going to be problematic since views are basically just aliases to queries. materializing this data into a proper snapshot table should be the solution. a snapshot row shouldnt ever change in the future once its calculated, so Id worry about a view where that might not be the case.

pkg/postgres/migrations/202512021500_stakerSharesView/up.go
```diff
@@ -0,0 +1,42 @@
...
+		select
+			coalesce(base.staker, wq.staker) as staker,
+			coalesce(base.strategy, wq.strategy) as strategy,
+			(coalesce(base.shares::numeric, 0) + coalesce(wq.shares::numeric, 0))::text as shares,
+			coalesce(base.snapshot, wq.snapshot) as snapshot
+		from staker_share_snapshots base
```
seanmcgary: Why not adjust the query that populates the `staker_share_snapshots` table to account for this case?

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -11,7 +11,7 @@ const operatorAllocationSnapshotsQuery = `
 		SELECT *,
 			   ROW_NUMBER() OVER (PARTITION BY operator, avs, strategy, operator_set_id, cast(block_time AS DATE) ORDER BY block_time DESC, log_index DESC) AS rn
 		FROM operator_allocations oa
-		INNER JOIN blocks b ON oa.block_number = b.number
+		INNER JOIN blocks b ON oa.effective_block = b.number
```
seanmcgary: How are you dealing with backwards compatibility?

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -42,8 +42,8 @@ const operatorAllocationSnapshotsQuery = `
 			-- Deallocation (decrease or no change): Round DOWN to current day
 			CASE
 				WHEN LAG(magnitude) OVER (PARTITION BY operator, avs, strategy, operator_set_id ORDER BY block_time, block_number, log_index) IS NULL THEN
-					-- First allocation: round down to current day (conservative default)
-					date_trunc('day', block_time)
+					-- First allocation: round up to next day
+					date_trunc('day', block_time) + INTERVAL '1' day
```
seanmcgary, line 51: How are you dealing with backwards compatibility?

pkg/rewards/withdrawalQueueShareSnapshots.go
seanmcgary: > A design principle that makes sense is that we should do as little rounding as possible

+1 to this. dont round, treat this table as a time-series table like the staker shares table

> I'd like us to ideally just insert into the staker_shares table.

Dont do this. The resulting values for the snapshots table should just query the staker shares and whatever other tables are needed to derive the values. Since we need to account for some kind of balances based on an effective timestamp, theres no clean way to trigger an insert for that, so querying it and deriving it is much cleaner and doesnt pollute the staker_shares table which is entirely sourced from on-chain events.

pkg/postgres/migrations/202511171438_withdrawalAndDeallocationQueues/up.go
ypatil12: Agreed. But you still need to make sure that the staker_shares table doesn't process withdrawals right when they're queued. So, you'd have both:

1. A queued withdrawal table that has the share decrement with the block 14 days from now. This is joined with staker shares
2. Staker shares doesn't have withdrawals added to it immediately

ypatil12:
My main point on the above design is that from a design perspective, we should strive to insert share increments/decrements into base bronze tables and then have the snapshot table process these share changes. It's cleaner than inserting a snapshotted record into an already snapshotted table... unless im missing something:

So... you have `withdrawal_queue_shares` which is joined with `staker_shares`, and that new table is eventually processed into `staker_share_snapshots`

pkg/rewards/withdrawalQueueShareSnapshots.go
ypatil12: Good point. We should just join a parallel `withdrawal_queue` table with `staker_shares` and then have that processed into `staker_share_snapshots`.

serichoi65:
```
STAKER REWARDS (Withdrawal Queue):
┌─────────────────────────────────────────────────┐
│ Bronze Tables (Blockchain Truth)               │
│ ├─> staker_shares                              │
│ └─> queued_slashing_withdrawals                │
│     (14-day window logic in query)             │
└─────────────────────────────────────────────────┘
              ↓
┌─────────────────────────────────────────────────┐
│ Snapshot Generation (stakerShareSnapshots.go)  │
│ • Queries BOTH bronze tables directly          │
│ • Applies 14-day window filter                 │
│ • Materializes into staker_share_snapshots     │
└─────────────────────────────────────────────────┘
              ↓
┌─────────────────────────────────────────────────┐
│ Final Snapshot (staker_share_snapshots)        │
│ • Includes base shares + withdrawal queue      │
│ • Ready for gold table consumption             │
└─────────────────────────────────────────────────┘

OPERATOR REWARDS (Deallocation Queue):
┌─────────────────────────────────────────────────┐
│ Bronze Tables (Blockchain Truth)               │
│ ├─> operator_shares                            │
│ └─> operator_allocations                       │
│     (COALESCE for backward compat)             │
└─────────────────────────────────────────────────┘
              ↓
┌─────────────────────────────────────────────────┐
│ Intermediate Snapshot                           │
│ • operator_allocation_snapshots                │
│   (14-day delay via effective_block)           │
└─────────────────────────────────────────────────┘
              ↓
┌─────────────────────────────────────────────────┐
│ Snapshot Generation (operatorShareSnapshots.go)│
│ • Queries operator_shares (bronze)             │
│ • Joins operator_allocation_snapshots          │
│ • Materializes into operator_share_snapshots   │
└─────────────────────────────────────────────────┘
              ↓
┌─────────────────────────────────────────────────┐
│ Final Snapshot (operator_share_snapshots)      │
│ • Includes base shares + allocations           │
│ • Ready for gold table consumption             │
└─────────────────────────────────────────────────┘
```

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: `COALESCE(oa.effective_block, oa.block_number)` handles old/new records now!

ypatil12: This file didn't exist before, so there shouldn't be a backwards compatibility issue?

pkg/rewards/stakerShareSnapshots.go
ypatil12, line 27: We should join the `withdrawal_queue_adjustments` here with `staker_shares`, and then have the snapshot calculation run

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -39,22 +39,51 @@ const stakerShareSnapshotsQuery = `
...
+		SELECT
+			staker,
+			strategy,
+			shares,
+			cast(day AS DATE) AS snapshot
+		FROM
+			cleaned_records
+		CROSS JOIN
+			generate_series(DATE(start_time), DATE(end_time) - interval '1' day, interval '1' day) AS day
+	),
```
ypatil12: I assume this will be done in a later commit, but just reminding that we also need to handle slashings while they're in the queue.

One of the edge cases is having a slashing in the queue. As part of this feature, shares are not decremented when they enter the queue. Thus, upon a slash we are decrementing from the shares as if the user *had never queued a withdrawal*. Then, once the withdrawal is completable, we do the actual decrement. Here is how this should work:

1. Alice has 200 shares and is delegated to Bob
2. Alice queues a withdrawal for 100 shares
    1. Sidecar State: Alice has 200 shares
    2. Sidecar Withdrawal Table: Alice has a withdrawal for 100 shares that will be completed 14 days from now. This table is joined with `staker_shares` in this snapshots table
3. Bob is slashed for 25%
    1. Sidecar State: Alice has 200 - 200*0.25 = 150 shares
    2. Sidecar Withdrawal Table: Alice’s withdrawal has 100 - 100 *0.25 = 75 shares 
        1. Here, when bob is slashed, we must search the withdrawal table for any withdrawals that are active for which Bob is the operator
4. Alices withdrawal is completable
    1. The staker_shares table should now properly decrement Alice’s shares to 150 - 75 = 75 shares

pkg/postgres/migrations/202511141700_withdrawalQueueAndAllocationRounding/up.go
```diff
@@ -14,14 +14,11 @@ type Migration struct {
...
 		// =============================================================================
-
-		// Add completion tracking (timestamps can be derived from blocks table via FK)
-		`alter table queued_slashing_withdrawals add column if not exists completion_block_number bigint`,
-
-		// Add FK constraint for completion block
-		`alter table queued_slashing_withdrawals add constraint fk_completion_block foreign key (completion_block_number) references blocks(number) on delete set null`,
+		// Note: Withdrawal queue logic uses withdrawable_date to determine when
+		// shares should stop earning rewards. The withdrawable_date is calculated as
+		// queued_date + 14 days. No additional columns needed in queued_slashing_withdrawals.
```
ypatil12, line 21: Don't we also need to do properly be backwards compatible with the `queued_slashing_withdrawals` table?

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -39,22 +39,51 @@ const stakerShareSnapshotsQuery = `
 	cleaned_records as (
 	  SELECT * FROM staker_share_windows
 	  WHERE start_time < end_time
+	),
```
ypatil12: We're re-adding the queued shares by joining. So it would be something like:

1. Alice has 200 shares
2. Alice queues a withdrawal for 50 shares
3. `queued_withdrawal_table` has 50 shares 

The staker_shares table then says Alice has 150 shares.

Once we join here, we do 150 - 50 again to get 100. We instead should do 150 + 50 to get 200.

pkg/rewards/operatorAllocationSnapshots.go
ypatil12: Is this an issue if `operatorAllocationSnapshots` didn't exist before?

ypatil12 (commented):
Withdrawal review

pkg/rewards/stakerShareSnapshots.go
serichoi65: This logic was already added. However, I added the new `queuedWithdrawalSlashingAdjustments` table that will be updated by slashingProcessor for a separation, and this also simplifies the query in stakerShareSnapshots.

pkg/rewards/stakerShareSnapshots.go
serichoi65: Yes, that's how it works currently from `(sr.shares::numeric + COALESCE(...))`
```
150 (on-chain shares) + 50 (queued shares) = 200 (effective shares for rewards) ✓
```

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
```diff
@@ -80,5 +80,171 @@ func (sp *SlashingProcessor) Process(blockNumber uint64, models map[string]types
...
+
+func (sp *SlashingProcessor) createSlashingAdjustments(slashEvent *SlashingEvent, blockNumber uint64) error {
+	// Find all active queued withdrawals for this operator/strategy
+	query := `
+		SELECT
+			qsw.staker,
+			qsw.strategy,
+			qsw.operator,
+			qsw.block_number as withdrawal_block_number,
+			? as slash_block_number,
```
seanmcgary: Can we use the `:namedTokens` rather that question marks? Makes it a LOT easier to see what values are getting inserted where and makes it so ordering doesnt matter.

seanmcgary (changes requested):

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
serichoi65: Updated!

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -25,7 +27,52 @@ const stakerShareSnapshotsQuery = `
...
+				 AND b_queued.block_time <= sr.block_time
+				 AND DATE(b_queued.block_time) + INTERVAL '14 days' > DATE '{{.snapshotDate}}'
+				 -- Backwards compatibility: only process records with valid data
+				 AND qsw.staker IS NOT NULL
+				 AND qsw.strategy IS NOT NULL
+				 AND qsw.operator IS NOT NULL
+				 AND qsw.shares_to_withdraw IS NOT NULL
+				 AND b_queued.block_time IS NOT NULL
+				), 0
+			))::text as shares,
```
ypatil12: Should be numeric?

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -25,7 +27,52 @@ const stakerShareSnapshotsQuery = `
...
+					COALESCE(qsw.shares_to_withdraw::numeric, 0) * COALESCE(
+						-- Get the latest (most recent) slash multiplier for this withdrawal
+						(SELECT adj.slash_multiplier::numeric
+						 FROM queued_withdrawal_slashing_adjustments adj
+						 WHERE adj.staker = qsw.staker
+						 AND adj.strategy = qsw.strategy
+						 AND adj.withdrawal_block_number = qsw.block_number
+						 -- Only consider slashing that happened before the cutoff date
+						 AND adj.slash_block_number <= (
+							 SELECT number FROM blocks WHERE block_time <= TIMESTAMP '{{.cutoffDate}}' ORDER BY number DESC LIMIT 1
```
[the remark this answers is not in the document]
ypatil12: We might need to add log_index to the `queued_slashing_withdrawals` table?

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
```diff
@@ -80,5 +80,169 @@ func (sp *SlashingProcessor) Process(blockNumber uint64, models map[string]types
...
+				(SELECT slash_multiplier
+				 FROM queued_withdrawal_slashing_adjustments adj
+				 WHERE adj.staker = qsw.staker
+				 AND adj.strategy = qsw.strategy
+				 AND adj.operator = qsw.operator
+				 AND adj.withdrawal_block_number = qsw.block_number
+				 ORDER BY adj.slash_block_number DESC
+				 LIMIT 1),
+				1
+			) * (1 - LEAST(@wadSlashed / 1e18, 0)) as slash_multiplier,
```
ypatil12: `LEAST(@wadSlashed / 1e18, 0)` would always return 0? Probably want `LEAST(@wadSlashed / 1e18, 1))`

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
```diff
@@ -80,5 +80,169 @@ func (sp *SlashingProcessor) Process(blockNumber uint64, models map[string]types
...
+			@blockNumber as block_number,
+			@transactionHash as transaction_hash,
+			@logIndex as log_index
+		FROM queued_slashing_withdrawals qsw
+		INNER JOIN blocks b_queued ON qsw.block_number = b_queued.number
+		WHERE qsw.operator = @operator
+		AND qsw.strategy = @strategy
+		-- Withdrawal was queued before this slash
+		AND qsw.block_number < @slashBlockNumber
```
ypatil12: We should check block number and log index. Since, a slash can happen right after withdrawal was queued

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -25,7 +27,52 @@ const stakerShareSnapshotsQuery = `
...
+				 AND qsw.strategy IS NOT NULL
+				 AND qsw.operator IS NOT NULL
+				 AND qsw.shares_to_withdraw IS NOT NULL
+				 AND b_queued.block_time IS NOT NULL
+				), 0
+			))::text as shares,
+			sr.snapshot_time
+		FROM snapshotted_records sr
+	),
+	-- Get the range for each staker, strategy pairing
```
ypatil12, line 95: So it seems like we're adding *both* the queued shares and completed shares and creating that as one entry? Correct me if I'm wrong. Ideally for a staker_share_snapshots table we'd want entries for each one of these:

0. Alice has 200 shares. At T0, Alice has 200 shares
1. Alice queues a withdrawal for 50 shares. At T1, Alice has 200 share
2. Alice is slashed for 25%. At T2 the `staker_shares` table has a -50 for Alice (37.5 come from `staker_shares` normally and 12.5 come from the queued withdrawal calc). Thus, `staker_snapshots` has 150 total shares for Alice. 
3. Alice is completable. At T3, Alice has 137.5 shares. 

We need to make sure that each of these is a unique entry into the snapshots table. Perhaps its time to just start testing these scenarios out

ypatil12: #2 in particular is important to ensure. It does appear that we do this, but just need to test :)

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -25,7 +27,52 @@ const stakerShareSnapshotsQuery = `
...
+						 FROM queued_withdrawal_slashing_adjustments adj
+						 WHERE adj.staker = qsw.staker
+						 AND adj.strategy = qsw.strategy
+						 AND adj.withdrawal_block_number = qsw.block_number
+						 -- Only consider slashing that happened before the cutoff date
+						 AND adj.slash_block_number <= (
+							 SELECT number FROM blocks WHERE block_time <= TIMESTAMP '{{.cutoffDate}}' ORDER BY number DESC LIMIT 1
+						 )
+						 ORDER BY adj.slash_block_number DESC
+						 LIMIT 1),
```
ypatil12: One thing we also need to properly handle is beacon chain slashes. Native ETH (eigenpod) shares can be slashed by both the beacon chain and an AVS. We can handle this after we get more traction on ensuring the other parts are correct, but want to call this out. In the `shares.go` eigenstate we do:

```golfing
	wadSlashed := big.NewInt(1e18)
	wadSlashed = wadSlashed.Mul(wadSlashed, new(big.Int).SetUint64(outputData.NewBeaconChainSlashingFactor))
	wadSlashed = wadSlashed.Div(wadSlashed, new(big.Int).SetUint64(outputData.PrevBeaconChainSlashingFactor))
	wadSlashed = wadSlashed.Sub(big.NewInt(1e18), wadSlashed)
```

ypatil12 (commented):
Accumulating the wads is smart! Brought up some minor cases, but generally we should be good to start really unit testing the `stakerShareSnapshots.go` file

pkg/rewards/stakerShareSnapshots.go
ypatil12: Will write down an equation for this tomorrow.

pkg/rewards/stakerShareSnapshots.go
serichoi65: Added the unit test about this, but one of the tests need to be uncommented when all changes are merged & deployed.

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
```diff
@@ -80,5 +80,172 @@ func (sp *SlashingProcessor) Process(blockNumber uint64, models map[string]types
...
+		INNER JOIN blocks b_queued ON qsw.block_number = b_queued.number
+		WHERE qsw.operator = @operator
+		AND qsw.strategy = @strategy
+		-- Withdrawal was queued before this slash (check block number AND log index for same-block events)
+		AND (
+			qsw.block_number < @slashBlockNumber
+			OR (qsw.block_number = @slashBlockNumber AND qsw.log_index < @logIndex)
+		)
+		-- Still within 14-day window (not yet completable)
+		AND DATE(b_queued.block_time) + INTERVAL '14 days' > (
```
ypatil12: The withdrawal queue window is 10 minutes on preprod/testnet but 14 days on mainnet. We'll have to think about how we should handle this on preprod/testnet environments

ypatil12: For now, should be good to merge this PR in, but I wrote it down as an open question to handle prior to deploying on preprod. My thinking here is to use the proper interval for each environment. However, tests (locally, integration) should assume we're on mainnet.

pkg/rewards/stakerShareSnapshots.go
ypatil12: I *believe* this case is already handled since beacon chain slashed propagate through decrements in `stakerShares`. What we need to ensure is that these types of slashes are not "double-decremented". 

We handles avs slashes through the pathway of `OperatorSlashed` event, which *should not* affect the `queued_slashing_withdrawals` table. Is there anywhere where we blankly assume a decrease in staker shares is a queued withdrawal?

Here is an integration test we should eventually have: 

T1: Alice has 200 shares
T2: Alice queues a withdrawal for 50 shares
	- EigenState: Alice has 150 shares
        - RewardsState: Alice has 200 shares
T3: Bob (Alice’s operator) is slashed for 25% for the BeaconChain Strategy. 
         - EigenState: -37.5 shares for Alice. Alice has 112.5 Shares
         - RewardsState: -50 shares for Alice. Alice has 150 shares
T4: Alice is slashed for 50% for on the BeaconChain
         - EigenState: -56.25 Shares for Alice. Alice has 56.25 shares
         - RewardsState: -75 shares for Alice. Alice has 75 shares
T15: Alice’s withdrawal is completable
          - EigenState: 56.25 shares
          - RewardsState: 56.25 shares

pkg/rewards/stakerShareSnapshots.go
```diff
@@ -25,7 +27,52 @@ const stakerShareSnapshotsQuery = `
 	 from ranked_staker_records
 	 where rn = 1
 	),
-	-- Get the range for each operator, strategy pairing
+	-- Join with withdrawal queue at bronze level before creating windows
```
ypatil12: We're still inserting these records *after* the snapshotting is being done. We ideally should insert them *before* the snapshotting is being done. Does doing so break the tests in any way?

Ideally, we only join with diffs and then we snapshot. Right now we're taking the staker share diffs, snapshotting, and then inserting these withdrawal queue adjustments

pkg/rewards/stakerShareSnapshots_test.go
```diff
@@ -157,3 +157,1001 @@ func Test_StakerShareSnapshots(t *testing.T) {
...
+			t.Logf("  [%d] Date: %s, Shares: %s", i, snap.Snapshot.Format(time.DateOnly), snap.Shares)
+		}
+
+		// Expected:
+		// T0: 100 shares
+		// T1: 100 shares (70 base + 30 queued)
+		// T2: 50 shares (35 base + 15 queued after 50% slash)
+		// T3: 35 shares (queued withdrawal no longer counts)
+
+		assert.GreaterOrEqual(t, len(snapshots), 3, "Should have at least 3 unique snapshots")
```
ypatil12: We should assert the actual values, not just the number of snapshots

ypatil12 (commented):
Some comments. If we need to close out this PR, wdyt about resolving the the non-testing comments and then merge this into a release branch (not master). Then, we can start a big testing push across all files. 

otherwise, probably best to keep this PR open and just start adding tests separately. I can start testing tmrw

pkg/rewards/stakerShareSnapshots.go
ypatil12: For backwards compatibility, we really should be doing if casing on the sabine fork block, like we do across the rewards calc for other hard forks

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
ypatil12: Note: these conversions don't work as intended because they convert queued.block_time into a date and then add 14 days. We need to add 14 days in *seconds* to the block_time

pkg/rewards/stakerShareSnapshots.go
serichoi65: Made a change to: Bronze diffs → Join → Snapshot windows → Expand

pkg/rewards/stakerShareSnapshots_test.go
serichoi65: Updated and applied to the whole test

pkg/rewards/stakerShareSnapshots.go
serichoi65: Applied the same block-based fork checking pattern like eigenState

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
serichoi65: Removed `DATE` and using `TIMESTAMP` to have block_time in seconds

PR #481, opened 2025-11-20
feat: implement Rewards v2.2 - Operator Set Rewards with Unique & Total Stake

serichoi65:
## Description

Implements Rewards v2.2 - Operator Set rewards with automatic stake-weighted distribution for both unique allocated stake and total delegated stake. This brings forward-looking reward capabilities (up to 2 years) to operator sets managed via AllocationManager, closing the architectural gap where only AVSDirectory-based AVSs could create forward-looking rewards.

### Withdrawal Queue Integration (Sabine Fork)
14-day withdrawal delay now integrated into staker share snapshots:

- Withdrawals queued on Day 1 → shares removed starting Day 15
- Slashing during queue period adjusts withdrawable amounts
- Tracked via queued_slashing_withdrawals table
- Migration: 202512101500_queuedWithdrawalSlashingAdjustments

### Unique Stake Rewards (Tables 15-17)
Rewards based on allocated unique stake to operator sets:

- Table 15: Operator rewards - calculates pro-rata distribution based on (shares × magnitude / max_magnitude) × multiplier
- Table 16: Staker rewards - distributes staker portion from Table 15, respects withdrawal queue
- Table 17: AVS refunds - refunds for operators without unique stake allocations

### Total Stake Rewards (Tables 18-20)
Rewards based on total delegated stake (operator set analog to Rewards v1):

- Table 18: Operator rewards - calculates pro-rata distribution based on shares × multiplier
- Table 19: Staker rewards - distributes staker portion from Table 18, queue-aware
- Table 20: AVS refunds - refunds for operators without sufficient total stake

### Refund Logic
Tables 17 and 20 calculate separate refund buckets (not subtractions) for three scenarios:

- Operators not registered to operator set
- Operators registered but without required stake (unique stake for 15-17, total stake for 18-20)
- Operators with stake but strategies not registered

These refunds are tracked separately and combined in final staging tables (Table 21).

### Slashable Period Tracking
Enhanced operatorSetOperatorRegistrationSnapshots to track slashable_until:

- Operators remain in snapshots until deregistration + 14 days
- Ensures correct reward distribution during post-deregistration slashable period
- Migration: 202512160000_addSlashableUntilToOperatorRegistrationSnapshots

## Type of change

## How Has This Been Tested?

Integration testing

## Checklist:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -907,13 +907,12 @@ func (c *Config) IsRewardsV2_2EnabledForCutoffDate(cutoffDate string) (bool, err
 	if err != nil {
 		return false, errors.Join(fmt.Errorf("failed to parse cutoff date %s", cutoffDate), err)
 	}
-	// TODO: Need to change to fork sabine
-	pecosForkDateTime, err := time.Parse(time.DateOnly, forks[RewardsFork_Pecos].Date)
+	sabineForkDateTime, err := time.Parse(time.DateOnly, forks[RewardsFork_Sabine].Date)
 	if err != nil {
-		return false, errors.Join(fmt.Errorf("failed to parse Pecos fork date %s", forks[RewardsFork_Pecos].Date), err)
+		return false, errors.Join(fmt.Errorf("failed to parse Sabine fork date %s", forks[RewardsFork_Sabine].Date), err)
 	}
 
-	return cutoffDateTime.Compare(pecosForkDateTime) >= 0, nil
+	return cutoffDateTime.Compare(sabineForkDateTime) >= 0, nil
 }
 
 // CanIgnoreIncorrectRewardsRoot returns true if the rewards root can be ignored for the given block number
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -118,6 +118,7 @@ type RewardsConfig struct {
 	GenerateStakerOperatorsTable bool
 	CalculateRewardsDaily        bool
 	WithdrawalQueueWindow        float64 // Duration in days for withdrawal queue period (14.0 for mainnet, 0.0069 for testnet/preprod ~10 min)
+	RewardsV2_2Enabled           bool
 }
 
 type StatsdConfig struct {
@@ -241,6 +242,7 @@ var (
 	RewardsGenerateStakerOperatorsTable = "rewards.generate_staker_operators_table"
 	RewardsCalculateRewardsDaily        = "rewards.calculate_rewards_daily"
 	RewardsWithdrawalQueueWindow        = "rewards.withdrawal_queue_window"
+	RewardsV2_2Enabled                  = "rewards.v2_2_enabled"
 
 	EthereumRpcBaseUrl               = "ethereum.rpc_url"
 	EthereumRpcContractCallBatchSize = "ethereum.contract_call_batch_size"
@@ -329,6 +331,7 @@ func NewConfig() *Config {
 			GenerateStakerOperatorsTable: viper.GetBool(normalizeFlagName(RewardsGenerateStakerOperatorsTable)),
 			CalculateRewardsDaily:        viper.GetBool(normalizeFlagName(RewardsCalculateRewardsDaily)),
 			WithdrawalQueueWindow:        FloatWithDefault(viper.GetFloat64(normalizeFlagName(RewardsWithdrawalQueueWindow)), getDefaultWithdrawalQueueDuration(chain)),
+			RewardsV2_2Enabled:           viper.GetBool(normalizeFlagName(RewardsV2_2Enabled)),
 		},
 
 		DataDogConfig: DataDogConfig{
@@ -893,6 +896,28 @@ func (c *Config) IsRewardsV2_1EnabledForCutoffDate(cutoffDate string) (bool, err
 	}
 
 	return cutoffDateTime.Compare(mississippiForkDateTime) >= 0, nil
+}
+
+func (c *Config) IsRewardsV2_2EnabledForCutoffDate(cutoffDate string) (bool, error) {
+	// Check global flag first - if disabled, return false regardless of date
+	if !c.Rewards.RewardsV2_2Enabled {
+		return false, nil
+	}
+
+	forks, err := c.GetRewardsSqlForkDates()
+	if err != nil {
+		return false, err
+	}
+	cutoffDateTime, err := time.Parse(time.DateOnly, cutoffDate)
+	if err != nil {
+		return false, errors.Join(fmt.Errorf("failed to parse cutoff date %s", cutoffDate), err)
+	}
+	pecosForkDateTime, err := time.Parse(time.DateOnly, forks[RewardsFork_Pecos].Date)
+	if err != nil {
+		return false, errors.Join(fmt.Errorf("failed to parse Pecos fork date %s", forks[RewardsFork_Pecos].Date), err)
+	}
+
+	return cutoffDateTime.Compare(pecosForkDateTime) >= 0, nil
 }
 
 // CanIgnoreIncorrectRewardsRoot returns true if the rewards root can be ignored for the given block number
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -899,11 +899,6 @@ func (c *Config) IsRewardsV2_1EnabledForCutoffDate(cutoffDate string) (bool, err
 }
 
 func (c *Config) IsRewardsV2_2EnabledForCutoffDate(cutoffDate string) (bool, error) {
-	// Check global flag first - if disabled, return false regardless of date
-	if !c.Rewards.RewardsV2_2Enabled {
-		return false, nil
-	}
-
 	forks, err := c.GetRewardsSqlForkDates()
 	if err != nil {
 		return false, err
@@ -912,6 +907,7 @@ func (c *Config) IsRewardsV2_2EnabledForCutoffDate(cutoffDate string) (bool, err
 	if err != nil {
 		return false, errors.Join(fmt.Errorf("failed to parse cutoff date %s", cutoffDate), err)
 	}
+	// TODO: Need to change to fork sabine
 	pecosForkDateTime, err := time.Parse(time.DateOnly, forks[RewardsFork_Pecos].Date)
 	if err != nil {
 		return false, errors.Join(fmt.Errorf("failed to parse Pecos fork date %s", forks[RewardsFork_Pecos].Date), err)
--- a/pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
+++ b/pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
@@ -2,6 +2,7 @@ package _202512160000_addSlashableUntilToOperatorRegistrationSnapshots
 
 import (
 	"database/sql"
+	"fmt"
 
 	"github.com/Layr-Labs/sidecar/internal/config"
 	"gorm.io/gorm"
@@ -11,20 +12,24 @@ type Migration struct {
 }
 
 func (m *Migration) Up(db *sql.DB, grm *gorm.DB, cfg *config.Config) error {
+	// Get the withdrawal queue window from config (14 days on mainnet, ~10 minutes on testnets)
+	withdrawalQueueWindow := cfg.Rewards.WithdrawalQueueWindow
+
 	queries := []string{
 		// Step 1: Add nullable column
 		`ALTER TABLE operator_set_operator_registration_snapshots
 		 ADD COLUMN IF NOT EXISTS slashable_until DATE`,
+	}
 
-		// Step 2: Backfill accurate slashable_until dates for deregistered operators
-		// NULL = operator is still active, DATE = deregistration_date + 14 days
-		`WITH deregistration_events AS (
+	// Step 2: Backfill accurate slashable_until dates for deregistered operators
+	// NULL = operator is still active, DATE = deregistration_date + withdrawal_queue_window
+	backfillQuery := fmt.Sprintf(`WITH deregistration_events AS (
 			SELECT
 				osor.operator,
 				osor.avs,
 				osor.operator_set_id,
 				DATE(b.block_time) as deregistration_date,
-				DATE(b.block_time) + INTERVAL '14 days' as slashable_until_date
+				DATE(b.block_time) + (%f * INTERVAL '1 day') as slashable_until_date
 			FROM operator_set_operator_registrations osor
 			JOIN blocks b ON osor.block_number = b.number
 			WHERE osor.is_active = FALSE
@@ -46,8 +51,9 @@ func (m *Migration) Up(db *sql.DB, grm *gorm.DB, cfg *config.Config) error {
 		WHERE osors.operator = ld.operator
 		  AND osors.avs = ld.avs
 		  AND osors.operator_set_id = ld.operator_set_id
-		  AND osors.snapshot <= ld.deregistration_date`,
-	}
+		  AND osors.snapshot <= ld.deregistration_date`, withdrawalQueueWindow)
+
+	queries = append(queries, backfillQuery)
 
 	for _, query := range queries {
 		if _, err := db.Exec(query); err != nil {
--- a/pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
+++ b/pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
@@ -1,0 +1,62 @@
+package _202512160000_addSlashableUntilToOperatorRegistrationSnapshots
+
+import (
+	"database/sql"
+
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"gorm.io/gorm"
+)
+
+type Migration struct {
+}
+
+func (m *Migration) Up(db *sql.DB, grm *gorm.DB, cfg *config.Config) error {
+	queries := []string{
+		// Step 1: Add nullable column
+		`ALTER TABLE operator_set_operator_registration_snapshots
+		 ADD COLUMN IF NOT EXISTS slashable_until DATE`,
+
+		// Step 2: Backfill accurate slashable_until dates for deregistered operators
+		// NULL = operator is still active, DATE = deregistration_date + 14 days
+		`WITH deregistration_events AS (
+			SELECT
+				osor.operator,
+				osor.avs,
+				osor.operator_set_id,
+				DATE(b.block_time) as deregistration_date,
+				DATE(b.block_time) + INTERVAL '14 days' as slashable_until_date
+			FROM operator_set_operator_registrations osor
+			JOIN blocks b ON osor.block_number = b.number
+			WHERE osor.is_active = FALSE
+		),
+		last_deregistration AS (
+			-- Get the most recent deregistration for each (operator, avs, operator_set_id)
+			SELECT DISTINCT ON (operator, avs, operator_set_id)
+				operator,
+				avs,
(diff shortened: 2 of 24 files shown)


serichoi65:
refer to this PR for previous comments: https://github.com/Layr-Labs/sidecar/pull/466

pkg/rewards/12_goldOperatorODOperatorSetRewardAmounts.go
```diff
@@ -78,6 +78,7 @@ func (rc *RewardsCalculator) GenerateGold12OperatorODOperatorSetRewardAmountsTab
 		rc.logger.Sugar().Infow("Rewards v2.1 is not enabled for this cutoff date, skipping GenerateGold12OperatorODOperatorSetRewardAmountsTable")
 		return nil
 	}
+
```
ypatil12: Nit: extra line

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,155 @@
...
+        rso.*,
+        oas.magnitude,
+        oas.max_magnitude,
+        oss.shares as operator_total_shares,
+        -- Calculate effective allocated stake for this strategy
+        CAST(
+            CAST(oss.shares AS DECIMAL(78,0)) *
+            CAST(oas.magnitude AS DECIMAL(78,0)) /
+            CAST(oas.max_magnitude AS DECIMAL(78,0))
+        AS VARCHAR) as allocated_stake
```
ypatil12: Why casting as VARCHAR if we later cast `to allocated_stake AS DECIMAL(78,0)` ?

pkg/rewards/16_goldStakerOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,133 @@
...
+        AND sd.snapshot = sss.snapshot
+    WHERE sss.shares > 0
+        AND sd.multiplier != 0
+),
+
+-- Step 4: Calculate each staker's weighted shares
+staker_weights AS (
+    SELECT
+        *,
+        CAST(shares AS DECIMAL(78,0)) * multiplier as staker_weight
```
ypatil12: Do we need to cast here? Don't see us casting `staker_weight` in older queries

pkg/rewards/16_goldStakerOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,133 @@
...
+        -- Staker's proportion of operator's total delegated stake
+        CASE
+            WHEN total_operator_weight > 0 THEN
+                staker_weight / total_operator_weight
+            ELSE 0
+        END as staker_proportion,
+        -- Staker's reward = proportion * operator's staker_split_total
+        CASE
+            WHEN total_operator_weight > 0 THEN
+                FLOOR((staker_weight / total_operator_weight) * staker_split_total)
```
ypatil12: Looks like we're rounding these calls to 15 decimal places:

```sql
        FLOOR((staker_weight / total_weight) * 1000000000000000) / 1000000000000000 AS staker_proportion
```

pkg/rewards/17_goldAvsOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/Layr-Labs/sidecar/internal/config"
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const _17_goldAvsOperatorSetUniqueStakeRewardsQuery = `
+CREATE TABLE {{.destTableName}} AS
+
+-- Step 1: Get the rows where operators have not registered for the AVS or if the AVS does not exist
+WITH not_registered_operators AS (
```
ypatil12: I'm unsure if the logic has to be this complex here, as the criteria for refunds is different from that of RewardsV2. RewardsV2 directly pays out operators so we had to handle the case for operators not registered, but included in the rewards submission. Thus, we had to have `tokens_per_registered_snapshot_decimal`. 

In this case, we should be able to check if the tokens distributed for a given snapshot day are 0. In that case, we refund the days tokens back to the avs.

ypatil12 (commented):
First pass on unique stake reward

internal/config/config.go
```diff
@@ -841,6 +844,24 @@ func (c *Config) IsRewardsV2_1EnabledForCutoffDate(cutoffDate string) (bool, err
...
+func (c *Config) IsRewardsV2_2EnabledForCutoffDate(cutoffDate string) (bool, error) {
+	forks, err := c.GetRewardsSqlForkDates()
+	if err != nil {
+		return false, err
+	}
+	cutoffDateTime, err := time.Parse(time.DateOnly, cutoffDate)
+	if err != nil {
+		return false, errors.Join(fmt.Errorf("failed to parse cutoff date %s", cutoffDate), err)
+	}
+	// TODO: Need to change to fork sabine
```
seanmcgary: Does this need to be changed now or at some point in the future?

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,153 @@
...
+operators_with_allocated_stake AS (
+    SELECT
+        rso.*,
+        oas.magnitude,
+        oas.max_magnitude,
+        oss.shares as operator_total_shares,
+        -- Calculate effective allocated stake for this strategy
+        CAST(oss.shares AS DECIMAL(78,0)) *
+        CAST(oas.magnitude AS DECIMAL(78,0)) /
+        CAST(oas.max_magnitude AS DECIMAL(78,0)) as allocated_stake
```
seanmcgary: pretty sure these should be `NUMERIC()` and not `DECIMAL()`. DECIMAL is what we used a LONG time ago and moved away from

seanmcgary (changes requested):

internal/config/config.go
serichoi65: This will be updated after we merge this PR https://github.com/Layr-Labs/sidecar/pull/474

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: Updated.

pkg/rewards/17_goldAvsOperatorSetUniqueStakeRewards.go
ypatil12: cc @0xrajath on this one

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,143 @@
...
+		la.avs,
+		la.operator_set_id,
+		cast(@cutoffDate as date) as snapshot
+	FROM latest_allocations la
+	LEFT JOIN latest_max_magnitudes lmm
+		ON la.operator = lmm.operator
+		AND la.strategy = lmm.strategy
+		AND lmm.rn = 1
+	WHERE la.rn = 1
+		AND la.magnitude > 0  -- Only include active allocations
```
ypatil12: Curious why we're combining max magnitudes and allocations into one table here? Generally practice has been to separate out each data type. Max Magnitudes and allocations will update in separate events

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,153 @@
...
+        ap.reward_submission_date
+    FROM {{.activeODRewardsTable}} ap
+    JOIN operator_set_operator_registration_snapshots osor
+        ON ap.avs = osor.avs
+       AND ap.operator_set_id = osor.operator_set_id
+       AND ap.snapshot = osor.snapshot
+       AND ap.operator = osor.operator
+),
+
+operators_with_allocated_stake AS (
```
ypatil12, line 38: We should also check that the operator is registered to the operatorSet here. If the operator is not registered, then they should not earn rewards. Note that deregistrations have a 14 day queue (like deallocations). 

The `OperatorRemovedFromOperatorSet` event emits the operator and operatorSet. However, the operator is still slashable for 14 days here. Likely need to create a rewards only registration table to handle this (similar to the shares table)?

ypatil12: We could update the event to add a `slashableUntil` parameter if that helps here. Note that for total stake rewards (and the other opSet rewards types) they should immediately stop earning for a deregistration but for slashable stake we need to wait 14 days.

internal/config/config.go
serichoi65: Updated to Fork_Sabine

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: - Added a new column `slashable_until` to track 14 day queue after deregistration occurs
- migration file backfills all previous registration snapshot rows -> do you think this is an overkill? i don't think this will take long too do

pkg/rewards/17_goldAvsOperatorSetUniqueStakeRewards.go
serichoi65: Simplified the refund logic

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: In the gold table, we need both max magnitudes and allocations together, so separating adds more complexity of adding more tables and unnecessarily complex query. We do ensure separating on the event level, but I think some snapshots has lightly combined events before.

pkg/postgres/migrations/202512161200_addMaxMagnitudeToAllocationSnapshots/up.go
```diff
@@ -0,0 +1,38 @@
...
+		// This combines allocation magnitude with max magnitude in the snapshot
+		// following the pattern of operator_share_snapshots and staker_share_snapshots
+		// which combine data from multiple sources that are consumed together in calculations
+		`ALTER TABLE operator_allocation_snapshots
+		 ADD COLUMN IF NOT EXISTS max_magnitude NUMERIC NOT NULL DEFAULT 0`,
+
+		// Create index for max_magnitude queries for performance
+		`CREATE INDEX IF NOT EXISTS idx_operator_allocation_snapshots_max_magnitude
+		 ON operator_allocation_snapshots(operator, strategy, snapshot)
+		 WHERE max_magnitude > 0`,
```
ypatil12, line 25: Max magnitude can theoretically go down to 0 for an operator. Is it fine if the migration checks on this?

pkg/rewards/operatorSetOperatorRegistrationSnapshots.go
```diff
@@ -75,6 +75,20 @@ marked_statuses AS (
 		-- Mark the next_block_time as the end_time for the range
 		-- Use coalesce because if the next_block_time for a registration is not closed, then we use cutoff_date
 		COALESCE(next_block_time, '{{.cutoffDate}}')::timestamp AS end_time,
+		-- Calculate slashable_until based on deregistration
```
ypatil12, line 78: We probably have to deal with backwards compatibility here?

pkg/rewards/operatorSetOperatorRegistrationSnapshots.go
```diff
@@ -75,6 +75,20 @@ marked_statuses AS (
...
+		-- NULL = still active (no deregistration)
+		CASE
+			WHEN next_is_active = FALSE THEN
+				-- Deregistration with 14-day slashability period
+				DATE(COALESCE(next_block_time, '{{.cutoffDate}}')::timestamp) + INTERVAL '14 days'
+			WHEN next_is_active IS NULL THEN
+				-- Still active (no deregistration event): NULL
+				NULL
+			ELSE
+				-- Re-registered: use next event time
```
ypatil12: A bit confused on the use case for re-registered? Is this when an operator deregisters then registers again? Is there a case when next_is_active is not false or null? How is that triggered?

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 const operatorAllocationSnapshotsQuery = `
-	insert into operator_allocation_snapshots(operator, avs, strategy, operator_set_id, magnitude, snapshot)
+	insert into operator_allocation_snapshots(operator, avs, strategy, operator_set_id, magnitude, max_magnitude, snapshot)
```
ypatil12, line 9: Noting that since this table itself is net-new as part of this feature, no need to add hard fork if cases

pkg/rewards/operatorAllocationSnapshots.go
ypatil12: Wait, I thought this entire table was post sabine???

pkg/rewards/operatorAllocationSnapshots.go
ypatil12: Im trying to remember which does but I don't think we've combined any snapshot events before, aside from delegation/undelegation and registration/deregistration and staker shares. Those *need* to be combined because those events are effectively pairs. Magnitude and allocations aren't though. When we snapshot the above, we take into account time series data on each event and then combine into a snapshot. 

What we're doing here is creating snapshots and then joining these tables together. I'm a bit worried by the final join, because the allocations can always change irrespective of the max magnitude changing.

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
ypatil12: That should be fine. One thing we will have to document is that we are adding 14 day in times but the contracts add 14 days in blocks (we do this for the withdrawal queue too). Blocks can be missed onchain but time is continuous. It's a fine edge case.

pkg/rewards/operatorAllocationSnapshots.go
ypatil12: Thought about this some more. One con of a separate table is state bloat. Max magnitudes update infrequently. However, we do need extensive test coverage on magnitudes not being affected by allocations/deallocation

ypatil12: Also, `maxMagnitude` is initialized at 1e18 in the contracts and is only decreasing. In our queries we assume is is 0 if there is no event, which we shouldn't do

pkg/rewards/operatorSetOperatorRegistrationSnapshots.go
serichoi65: Backwards compatibility is already handled! When adding the `slashable_until` column, the migration backfills `slashable_until` by recalculating from historical deregistration events in the `operator_set_operator_registrations` table

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: Yes, updating `maxMagnitude` default to 1e18 when no MagnitudeUpdated event exists, matching contract initialization.

pkg/rewards/operatorSetOperatorRegistrationSnapshots.go
serichoi65: I was initially thinking of deregister then register again, but that case is correctly covered. Removing ELSE clause.
Now:
- next_is_active = FALSE → Deregistration with 14-day slashability
- next_is_active IS NULL → Still active (no deregistration)

pkg/postgres/migrations/202512161200_addMaxMagnitudeToAllocationSnapshots/up.go
serichoi65: Yes, the partial index is fine as `max_magnitude = 0` is always filtered out by the query in gold table 15 (line 60). Operators with zero capacity correctly receive no rewards, whether they're in the index or not.

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: Noted and updated!

pkg/rewards/operatorAllocationSnapshots.go
serichoi65: I think keeping combined as is is the better approach, so I agree that we need an extensive test coverage to validate its correctness.

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,213 @@
...
+             AND dop.snapshot <= dop.slashable_until
+            THEN TRUE
+            ELSE FALSE
+        END as in_deregistration_queue
+    FROM distinct_operators dop
+),
+
+-- Step 4: Calculate cumulative slash multiplier during deregistration queue
+-- Slashing only affects rewards during the 14-day deregistration period
+operators_with_slash_multiplier AS (
```
ypatil12, line 118: What is the purpose of the slash multiplier here?

pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
```diff
@@ -0,0 +1,62 @@
...
+
+		// Step 2: Backfill accurate slashable_until dates for deregistered operators
+		// NULL = operator is still active, DATE = deregistration_date + 14 days
+		`WITH deregistration_events AS (
+			SELECT
+				osor.operator,
+				osor.avs,
+				osor.operator_set_id,
+				DATE(b.block_time) as deregistration_date,
+				DATE(b.block_time) + INTERVAL '14 days' as slashable_until_date
```
ypatil12: `slashable_until_date` is different on mainnet and other envs

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
ypatil12: If a slash has occurred, it should already be reflected in the operator's shares. All we need to do is verify that the operator is still slashable (ie. registered OR slashable_until not hit)

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,213 @@
+package rewards
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
```
ypatil12, line 7: We also need to add the events to the EigenState so they are indexed. See `OperatorDirectedOperatorSetRewardsSubmissionCreated` for an example

pkg/postgres/migrations/202512160000_addSlashableUntilToOperatorRegistrationSnapshots/up.go
serichoi65: I already had the config values - updated to use that as variable instead of a hardcoded value.

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: They are already indexed! Check `pkg/eigenState/eigenState.go`

pkg/rewards/15_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: slash multiplier is ONLY for the deregistration queue period. It handles slashing that occurs after deregistration but before the operator fully exits

PR #494, opened 2026-01-22
fix: rewards v2.2 audit fixes

serichoi65:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings


pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
```diff
@@ -173,8 +185,9 @@ func (sp *SlashingProcessor) createSlashingAdjustments(slashEvent *SlashingEvent
 			@logIndex as log_index
 		FROM queued_slashing_withdrawals qsw
```
ypatil12, line 198: So we're selecting from `queued_slashing_withdrawals`, but that hasn't been committed yet. Thus, we could have a lag for slashes that occur in the same block as a queued withdrawal. The reason the tests pass is that they insert the slash into the queued withdrawal table (fine for tests, but doesn't match how this works E2E). 

Comment from other PR:

> Since this query is selecting from queued_slashing_withdrawal table, which hasn't been committed yet (ie. latest block not updated), could we have a staleness issue? Edge case: a withdrawal has been queued in this block and then we are slashed a couple times in the same block. 

We'll need to get the accumulator for withdrawals that occurred in this given block

ypatil12: For a regression, what we can do is something like:

```go
return sharesModel.HandleStateChange(&beaconChainSlashingFactorDecreasedLog)
```

Since the handleStateChange is called prior to committing the final state to the DB.

pkg/eigenState/precommitProcessors/slashingProcessor/slashing_test.go
```diff
@@ -300,3 +302,794 @@ func processSlashing(stakerSharesModel *stakerShares.StakerSharesModel, allocati
...
+		require.NoError(t, res.Error, fmt.Sprintf("Failed to insert block %d", block.number))
+	}
+}
+
+func insertQueuedWithdrawal(t *testing.T, grm *gorm.DB, staker, operator, strategy string, blockNumber uint64, shares string) {
+	insertQueuedWithdrawalWithLogIndex(t, grm, staker, operator, strategy, blockNumber, 1, shares)
+}
+
+func insertQueuedWithdrawalWithLogIndex(t *testing.T, grm *gorm.DB, staker, operator, strategy string, blockNumber uint64, logIndex uint64, shares string) {
+	res := grm.Exec(`
```
ypatil12, line 1134: I realized inserting this directly is causing us to miss the edge case of a withdrawal and a slash in the same block. In this case, `queued_slashing_withdrawals` won't have its state committed yet and db won't be up to date. For this edge case of a withdrawal and slash in the same block, we should do something like:

```go
	return sharesModel.HandleStateChange(&beaconChainSlashingFactorDecreasedLog)
```

And then use the accumulator in `slashing.go`

pkg/eigenState/precommitProcessors/slashingProcessor/slashing_test.go
```diff
@@ -300,3 +302,794 @@ func processSlashing(stakerSharesModel *stakerShares.StakerSharesModel, allocati
...
+		`, "0xstaker2", "0xstrategy2", 1010).Scan(&multiplier)
+		require.NoError(t, res.Error)
+		assert.Contains(t, multiplier, "0.375", "Expected cumulative multiplier 0.375 (0.75 * 0.5)")
+	})
+
+	// CSA-3: Same-block event ordering (log_index precedence)
+	t.Run("CSA-3: Same-block event ordering", func(t *testing.T) {
+		cleanupCSATest(t, grm)
+
+		// Setup: Staker queues withdrawal on block 1000, log_index 2
```
ypatil12, line 467: Ordering reversed here, we want to queue withdrawal (early log index) then slash (later log index) per test plan

ypatil12: This is the edge case for the accumulator

pkg/eigenState/totalStakeRewardSubmissions/totalStakeRewardSubmissions.go
```diff
@@ -0,0 +1,451 @@
...
+			Logger: logger,
+		},
+		DB:               grm,
+		logger:           logger,
+		globalConfig:     globalConfig,
+		stateAccumulator: make(map[uint64]map[types.SlotID]*TotalStakeRewardSubmission),
+		committedState:   make(map[uint64][]*TotalStakeRewardSubmission),
+	}
+
+	esm.RegisterState(model, 13)
```
elhajin: the index 13 is used for `OperatorSetOperatorRegistrationModel` .. we should use 25

pkg/eigenState/uniqueStakeRewardSubmissions/uniqueStakeRewardSubmissions.go
```diff
@@ -0,0 +1,451 @@
...
+			Logger: logger,
+		},
+		DB:               grm,
+		logger:           logger,
+		globalConfig:     globalConfig,
+		stateAccumulator: make(map[uint64]map[types.SlotID]*UniqueStakeRewardSubmission),
+		committedState:   make(map[uint64][]*UniqueStakeRewardSubmission),
+	}
+
+	esm.RegisterState(model, 12)
```
elhajin: index 12 is used for `OperatorSetSplitModel` .. should be 24

elhajin (changes requested):

pkg/rewards/rewards.go
elhajin: why we call this again ? the `OperatorAllocationSnapshot` is already inserted as we called that before this gonna be a noop

pkg/eigenState/precommitProcessors/slashingProcessor/slashing.go
ypatil12: Just flagging this @serichoi65

pkg/rewards/15_goldActiveUniqueAndTotalStakeRewards.go
```diff
@@ -0,0 +1,180 @@
...
+),
+
+-- Step 6: Calculate tokens per day
+-- For stake-based rewards, tokens_per_day is the daily amount to be distributed pro-rata
+tokens_per_day AS (
+    SELECT
+        *,
+        -- Total tokens per day = floor(amount / duration_in_days)
+        -- duration is in seconds, convert to days
+        FLOOR(CAST(amount AS NUMERIC(78,0)) / (duration / 86400)) AS tokens_per_day_decimal
```
ypatil12: Let's stay away from casting to `numeric(78,0)`. The *old* docs say we do, but we do not do so in the code currently. 

amount is already of type numeric, so we can simply just do `FLOOR(amount/(duration/86400))`

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,291 @@
...
+
+import (
+	"github.com/Layr-Labs/sidecar/pkg/rewardsUtils"
+	"go.uber.org/zap"
+)
+
+const _16_goldOperatorOperatorSetUniqueStakeRewardsQuery = `
+CREATE TABLE {{.destTableName}} AS
+
+-- ============================================================================
```
ypatil12, line 11: Part 1 here doesn't make sense - we are taking active rewards from table 11, but it should be taken from table 15. This query should start from step 2 (it seems like we're doing the same thing in step 1 and step 2)

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,291 @@
...
+       AND ap.snapshot >= osor.snapshot
+       AND (osor.slashable_until IS NULL OR ap.snapshot <= osor.slashable_until)
+),
+
+-- OD Step 2: Calculate allocated weight per operator
+od_operators_with_weight AS (
+    SELECT
+        rso.*,
+        SUM(
+            CAST(oss.shares AS NUMERIC(78,0)) *
```
ypatil12: Let's stay away from casting as numeric(78,0). Type is already numeric

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,291 @@
...
+
+-- Stake Step 4: Calculate pro-rata tokens per operator and prepare for common processing
+stake_distinct_operators AS (
+    SELECT
+        pow.reward_hash, pow.snapshot, pow.token,
+        FLOOR(pow.tokens_per_day_decimal * pow.operator_allocated_weight / ptw.total_weight) as tokens_per_registered_snapshot_decimal,
+        pow.avs, pow.operator_set_id, pow.operator, pow.strategy, pow.multiplier,
+        pow.reward_submission_date, pow.registration_snapshot, pow.slashable_until,
+        pow.operator_allocated_weight,
+        'stake' as reward_source
```
ypatil12: What is the purpose of reward_source?

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,291 @@
...
+            WHEN slashable_until IS NOT NULL
+             AND snapshot > (slashable_until - INTERVAL '14 days')
+             AND snapshot <= slashable_until
+            THEN TRUE
+            ELSE FALSE
+        END as in_deregistration_queue
+    FROM combined_operators
+),
+
+-- Common Step 2: Calculate cumulative slash multiplier during deregistration queue
```
ypatil12: When an operator is slashed, we already decrement their shares. The deregistration queue doesn't need to be factored here - this should already be taken care of in operator registration snapshots (via the `slashable_until` parameter). 

The key issue to solve for is not accounting for queued withdrawals for an operator: 

1. Operator has a staker, Alice. Alice has 500 shares
2. Alice queues a withdrawal for 200 shares. The operator now has 300 shares. But, we want the operator to have 500 shares!

How do we ensure that the queued withdrawal for Alice is counted in the operator shares? Here is the discrepancy: Operator shares are decremented *immediately* when a withdrawal is queued. Therefore, the number of shares an operator has does not handle the queued withdrawal adjustments. To fix, one option there is to add these shares back into operator_share_snapshots, like we do for staker shares.

ypatil12 (commented):
Notes on table 16

pkg/eigenState/precommitProcessors/slashingProcessor/slashing_test.go
serichoi65: Added as a new test case.

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: Removed.

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
serichoi65: Irrelevant after removing Part 1.

pkg/eigenState/queuedSlashingWithdrawals/queuedSlashingWithdrawals.go
elhajin, line 127: Strategy is stored without `strings.ToLower()` here, the slashing adjustment query joins on `qsw.strategy = @strategy` ,if the original event had mixed-case, the join silently fails and no adjustments are created.

should lowercase the strategy here

PR #500, opened 2026-02-04
feat: implement rewards v2.2 - Operator Set Rewards with Unique & Total Stake

serichoi65:
## Description

Implements Rewards v2.2 - Operator Set rewards with automatic stake-weighted distribution for both unique allocated stake and total delegated stake. This brings forward-looking reward capabilities (up to 2 years) to operator sets managed via AllocationManager.

**Key tables added:**
- Table 15: Active unique and total stake rewards
- Table 16-18: Unique stake rewards (operator → staker → AVS refunds)
- Table 19-21: Total stake rewards (operator → staker → AVS refunds)

## Type of change

## How Has This Been Tested?

`Test_RewardsV2_2` passes.

## Checklist:


pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,225 @@
...
+        ap.multiplier,
+        ap.reward_submission_date,
+        osor.operator,
+        osor.snapshot as registration_snapshot,
+        osor.slashable_until
+    FROM {{.activeStakeRewardsTable}} ap
+    JOIN operator_set_operator_registration_snapshots osor
+        ON ap.avs = osor.avs
+       AND ap.operator_set_id = osor.operator_set_id
+       AND ap.snapshot >= osor.snapshot
```
elhajin: This should be `=`. Registration snapshots already have one row per day — using `>=` matches all previous days, so the weight gets summed N times in the window function. The `>=` was a workaround to keep deregistered operators in the queue, but we don't need that for pt-1.

Also drop the `slashable_until` filter bellow it's not needed with `=`

pkg/rewards/16_goldOperatorOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,225 @@
...
+-- ============================================================================
+
+-- Step 5: Check deregistration queue status
+operators_with_deregistration_status AS (
+    SELECT
+        *,
+        CASE
+            WHEN slashable_until IS NOT NULL
+             AND snapshot > (slashable_until - INTERVAL '14 days')
+             AND snapshot <= slashable_until
```
elhajin: steps 5, 6,7 - can be removed entirely which will fix all bellow issues : 

1. **slash_multiplier double-counts** : when a slash happens the contract emits `AllocationUpdated`, `MaxMagnitudeUpdated`, and `OperatorSharesDecreased`, so `shares`, `magnitude`, and `maxMagnitude` in the snapshots already reflect it. Applying `slash_multiplier` on top reduces the operator's tokens again (applies slashing twice)

2. **The deregistration queue is pt-2.** There are two types of "deregistration": full deregister (`deregisterFromOperatorSets`) and magnitude decrease (`modifyAllocations`). For magnitude decrease, `effective_block` in allocation snapshots already handles the delay, old magnitude persists until `DEALLOCATION_DELAY` passes, so the operator keeps earning during that period automatically. For full deregister, the operator stays slashable for `DEALLOCATION_DELAY` but earning during that window is a pt-2 feature (alongside staker queue rewards). For pt-1: deregister = stop earning, deallocate = `effective_block` handles it.

3. **Hardcoded 14 days** on line 110 — the registration snapshot generator uses `{{.withdrawalQueueWindow}}` which is configurable per chain (14 days on mainnet, ~10 min on testnets), but this query hardcodes `INTERVAL '14 days'`. Would break on testnets.

After removing, rewire `operator_splits` to read from `distinct_operators` directly.

pkg/rewards/18_goldAvsOperatorSetUniqueStakeRewards.go
```diff
@@ -0,0 +1,111 @@
...
+-- Step 1: Calculate total tokens available per (reward_hash, snapshot, avs, operator_set_id)
+-- Table 15 has pool-level data without operator column
+WITH total_available_tokens AS (
+    SELECT
+        reward_hash,
+        snapshot,
+        token,
+        avs,
+        operator_set_id,
+        SUM(tokens_per_day_decimal) as total_tokens
```
elhajin: `tokens_per_day_decimal` is the same value across all strategies for a given (`reward_hash` , `snapshot`, `avs`, `operator_set_id`) , it's `floor(amount / duration)` where `amount` is the total reward, duplicated per strategy in `unique_stake_reward_submissions`. The GROUP BY doesn't include `strategy`, so `SUM` inflates `total_tokens` by N (number of strategies).

Same issue in Table 21 for total_stake. (this is exploitable)


Fix: `MAX(tokens_per_day_decimal)` instead of `SUM` , so `MAX` just picks the one value.

pkg/rewards/20_goldStakerOperatorSetTotalStakeRewards.go
```diff
@@ -0,0 +1,145 @@
...
+        SUM(CAST(sss.shares AS NUMERIC(78,0))) as total_shares
+    FROM staker_delegations sd
+    JOIN operator_set_strategy_registration_snapshots ossr
+        ON sd.avs = ossr.avs
+        AND sd.operator_set_id = ossr.operator_set_id
+        AND sd.snapshot = ossr.snapshot
+    JOIN staker_share_snapshots sss
+        ON sd.staker = sss.staker
+        AND ossr.strategy = sss.strategy
+        AND sd.snapshot = sss.snapshot
```
elhajin: two issues, same root cause ,  this CTE joins through `operator_set_strategy_registration_snapshots` instead of `activeStakeRewardsTable` (Table 15).

1. **Wrong strategy set**: `ossr` gives ALL strategies registered in the operator set. Should only include strategies from the reward submission. If the operator set has strategies {S1, S2, S3} but the reward only covers {S1, S2}, staker shares in S3 are incorrectly included in the weight.

2. **Missing multiplier**: Without Table 15, there's no `multiplier` column. Table 19 uses `SUM(shares * multiplier)` for operator pro-rata, Table 17 uses `SUM(shares * multiplier)` for unique_stake staker weight, but this sums raw shares. Stakers in high-multiplier strategies get the same intra-operator weight as low-multiplier ones.

**Fix**: replace `operator_set_strategy_registration_snapshots` with `{{.activeStakeRewardsTable}}` (filtered `reward_type = 'total_stake'`), join on `reward_hash, avs, operator_set_id, snapshot`, use `asr.strategy` for the staker_share_snapshots join, and `SUM(shares * asr.multiplier)` for staker weight — same pattern as Table 17

pkg/rewards/operatorAllocationSnapshots.go
```diff
@@ -0,0 +1,187 @@
...
+	),
+	ranked_max_magnitude_records as (
+		SELECT
+			omm.operator,
+			omm.strategy,
+			omm.max_magnitude,
+			b.block_time,
+			omm.block_number,
+			omm.log_index,
+			ROW_NUMBER() OVER (PARTITION BY omm.operator, omm.strategy, cast(b.block_time AS DATE) ORDER BY b.block_time DESC, omm.log_index DESC) AS rn
```
elhajin: NIT : Inconsistent ordering with line 12. Both can use `ORDER BY block_number DESC, log_index DESC` only simpler and unambiguous. `block_time` is redundant since it's derived from `block_number`

pkg/rewards/operatorShareSnapshots.go
```diff
@@ -30,31 +30,58 @@ operator_share_windows as (
...
+	SELECT
+		coalesce(base.operator, alloc.operator) as operator,
+		coalesce(base.strategy, alloc.strategy) as strategy,
+		coalesce(alloc.total_magnitude, base.shares::numeric) as shares,
+		coalesce(base.snapshot, alloc.snapshot) as snapshot
+	FROM base_snapshots base
+	FULL OUTER JOIN allocation_adjustments alloc
+		ON base.operator = alloc.operator
+		AND base.strategy = alloc.strategy
+		AND base.snapshot = alloc.snapshot
```
elhajin: Dead code , `operator_allocation_snapshots` is populated  after `operatorShareSnapshots` ,the CTE always returns 0 rows, so `coalesce(alloc.total_magnitude, base.shares)` always falls through to `base.shares`. Should be removed 

- beyond the execution order issue, `SUM(magnitude)` grouped by `(operator, strategy, snapshot)` is not a valid quantity. Magnitudes are per-operator-set allocations percentage, summing them across sets has no meaning and would corrupt `shares` if this code ever ran.

**Fix** : remove both `allocation_adjustments` and `combined_snapshots` CTEs , and select from `base_snapshot`

pkg/rewards/19_goldOperatorOperatorSetTotalStakeRewards.go
```diff
@@ -0,0 +1,139 @@
...
+       AND ap.snapshot = osor.snapshot
+    WHERE ap.reward_type = 'total_stake'
+),
+
+-- Step 2: Calculate weighted total stake per operator
+operators_with_weight AS (
+    SELECT
+        rso.*,
+        oss.shares as operator_total_shares,
+        SUM(CAST(oss.shares AS NUMERIC(78,0)) * rso.multiplier) OVER (
```
elhajin: NIT : `shares` are already `numeric` columns. The `CAST(... AS NUMERIC(78,0))` is a no-op (same for `amount`

elhajin (commented):
what should be addressed :

- **T16 — drop Steps 5-7 (deregistration queue + slash multiplier).** The allocation snapshots already handle deallocation delay via `effective_block` and slashing via live magnitude/shares updates. The queue logic double-counts slashing and the `>=` join inflates operator weight. Simplifies the query from 8 CTEs to 5. Deregistered-operator rewards can be pt-2.
- **T18/T21 — `SUM(tokens_per_day_decimal)` → `MAX`.** `tokens_per_day_decimal` is duplicated per strategy; the GROUP BY doesn't include `strategy`, so SUM inflates `total_tokens` by N×.
- **T20 — join through Table 15 instead of `operator_set_strategy_registration_snapshots`.** Current join pulls all strategies in the operator set (not just reward submission strategies) and drops multipliers. Should mirror T17's pattern.
- **`operatorShareSnapshots.go` — remove `allocation_adjustments` CTE.** Dead code (table it reads isn't populated yet), and the `SUM(magnitude)` across operator sets is semantically invalid.

details in inline comments

pkg/rewards/20_goldStakerOperatorSetTotalStakeRewards.go
serichoi65: `reward_type = 'total_stake'` filtered in table 19.

pkg/rewards/18_goldAvsOperatorSetUniqueStakeRewards.go
```diff
@@ -17,7 +17,7 @@ WITH total_available_tokens AS (
         token,
         avs,
         operator_set_id,
-        SUM(tokens_per_day_decimal) as total_tokens
+        MAX(tokens_per_day_decimal) as total_tokens
     FROM {{.activeStakeRewardsTable}}
     GROUP BY reward_hash, snapshot, token, avs, operator_set_id
```
elhajin, line 23: missing `reward_type` filter in `total_available_tokens` 

Table 15 has both types. T18 checks distribution against T16/T17, which only contain `unique_stake rows` . T16/T17 will **never** have rows for `total_stake reward_hash` , so the *LEFT JOINs* always produce `operator_distributed = 0, staker_distributed = 0` for every `total_stake` hash, and the refund always triggers. 

Same in reverse: T21 always refunds every unique_stake hash (checks T19/T20 which never have unique_stake rows).

 every reward of either type gets double-paid 


**Fix** : add `WHERE reward_type = 'unique_stake'` to T18's `total_available_tokens` CTE, and `WHERE reward_type = 'total_stake'` to T21's.

cc @serichoi65

pkg/rewards/15_goldActiveUniqueAndTotalStakeRewards.go
```diff
@@ -0,0 +1,180 @@
...
+        strategy_index,
+        multiplier,
+        start_timestamp,
+        end_timestamp,
+        duration,
+        reward_type,
+        block_number,
+        block_time,
+        block_date,
+        CAST('{{.cutoffDate}}' AS TIMESTAMP(6)) AS global_end_inclusive
```
elhajin: - the `global_end_inclusive` will cause refund of full amount for every snapshot 

    - T15 generates daily snapshots up to `cutoffDate` , `cutoffdate` here is  included 
    - Registration snapshots use `end_time - 1 day`, so they only go up to `cutoffDate - 1`
    -  T16/T20 joins T15 with registrations thus no match for cutoffDate → 0 distributed
    - T18/T21 (refund tables) sees 0 distributed thus full refund to AVS for that snapshot
    - That day's rewards are never distributed and always returned to the avs 

This happens **every day** for every active `unique_stake` and `total_stake` reward


- **Fix** : 
```diff
// in table 15  in CTE *active_rewards_modified*
- CAST('{{.cutoffDate}}' AS TIMESTAMP(6)) AS global_end_inclusive
+ CAST('{{.cutoffDate}}' AS TIMESTAMP(6)) - interval '1' day AS global_end_inclusive
```

pkg/rewards/rewards.go
elhajin, line 528: - `unique_stake` and `total_stake` rewards submissions are excluded from merkleization , thus they won't be part of the root , and are not claimable on-chain 

**FIX** : should  include `reward_hash` values from `stake_operator_set_rewards`  to ` all_combined_rewards` 

cc @serichoi65

PR #510, opened 2026-03-11
fix: only include same-block delegated stakers whose operator matches…

elhajin:
## Description

Fixes cross-operator slashing bug where stakers delegating to an unrelated operator in the same block as a slash event were incorrectly included in the slashed set.

## Type of change

- [x] Chore (non-breaking change which does not add functionality)

## How Has This Been Tested?

- Added regression test `Should not slash unrelated same-block delegation` in slashingProcessor test suite
- Verified existing slashing tests still pass
- PoC test (now failing as expected) confirms victim staker is no longer incorrectly slashed

## Checklist:


serichoi65 (approved):
LGTM

